PR #21, opened 2016-06-07
Predispose translations and add italian language

fede91it:


tabacitu:
Looking good! Great job, thanks!

fede91it:
Do you know how I can use this specific commit with composer?

tabacitu:
Well you can specify a certain version in your composer.json's require array (say 0.6.1) and then run composer update. That way composer will only resolve the dependencies if that version is met.

PR #33, opened 2016-08-28
Make the admin routes serializable

SabatinoMasala:
Removed the closure-based redirect, so the routes can be cached (serialized) with `php artisan route:cache`.


tabacitu:
Thank you! :-)

SabatinoMasala:
Thanks for merging! Quick question though, why does StyleCI keep failing, while PHP Code Sniffer gives me no errors / warnings on AdminController.php (using PSR-2)?

tabacitu:
I think StyleCI is pickier than PHP Code Sniffer. This time, for example, it didn't like that [a comment of yours did not have a trailing dot](https://github.com/Laravel-Backpack/Base/commit/6b6205b9557316991616de57ef9292b711058315) :-)  

It's set to autocorrect, so after any merge StyleCI will do its thing. Not even worth looking if StyleCI approves before merging, really.

SabatinoMasala:
Ah I see! Thanks for elaborating! :)

PR #41, opened 2016-09-21
included a micro blockui library for better user feedback

OwenMelbz:
usage

jQuery(‘element’).blockui(); // blocks with a micro loader

jQuery(‘element’).unblockui(); //unblocks previous micro loader

### DIFF
--- a/src/resources/views/layout.blade.php
+++ b/src/resources/views/layout.blade.php
@@ -26,6 +26,7 @@
 
     <link rel="stylesheet" href="{{ asset('vendor/adminlte/') }}/plugins/pace/pace.min.css">
     <link rel="stylesheet" href="{{ asset('vendor/backpack/pnotify/pnotify.custom.min.css') }}">
+    <link rel="stylesheet" href="{{ asset('vendor/backpack/owenmelbz/blockui.min.css') }}">
 
     <!-- BackPack Base CSS -->
     <link rel="stylesheet" href="{{ asset('vendor/backpack/backpack.base.css') }}">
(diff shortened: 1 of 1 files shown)


tabacitu:
Hi @OwenMelbz ,

For this to work for others too, the css and js files also need to be placed inside Base's ```src/public``` folder, and then published.

But first: what does this do and how does it look?

Cheers!

OwenMelbz:
Its used within the image uploader tweaks, for CRUD, however the header template was controlled via base, so wasnt sure how best to include it?

If you checkout this and the image uploader branch tweaks, you'll see it in action when you pick a file from the uploader

lloy0076:
@tabacitu - I'd be inclined to rework this so that the files were under a more 'backpack' like place (nothing wrong with "owenmelbz" really but we really should try to keep things under a common namespace/file directory if we can).

Also, a *grumpy* reviewer might point out that this just wouldn't work at all because the published files aren't there at all 😈

tabacitu:
Well this has been sitting here long enough :-)  My worry then, and it's still a worry for me now, is that if this is always triggered when AJAXing, small AJAX calls, like details_row or AjaxDataTables might have show it for a split second - and it might prove to be annoying instead of helpful for such small interactions.

@OwenMelbz do you happen to have a few screenshots of how you made it look? What's your opinion on the above?

OwenMelbz:
I can find a demo to attach if need be.

However we achieve this, I think we basically just need a way to have a loading animation usable for the platform. For things such as image uploaders etc rather than the UI just sitting there.

I don't think we have to hook it into every Ajax request, but when and where appropriate to let users know things are happening.

We can take a different approach, this was just included as the image uploaded fixes/extensions benefited from it at the time

AbbyJanke:
@OwenMelbz is this still something of interest? There’s currently no CSS or JS in your commit so i can’t really test it. 

I know that in January Pace.ja was added maybe that takes place of this or am i understanding what this was doing?

OwenMelbz:
Erm no it does do different things.

It specifically blocked parts of the UI whilst something is happening - prevents double clicking a button for example.

The PR itself was complicated because it needed to be included in base - to be able to use it in crud at the time.

However it’s so outdated now this PR isn’t likely applicable anymore

AbbyJanke:
Alrighty, was just trying to help clean up older and outdated issues/prs.

tabacitu:
Thank you @OwenMelbz & @AbbyJanke !

PR #43, opened 2016-09-27
added functionality needed for enqueuing

OwenMelbz:
Need guidance how best to actually implement this functionality.

**usage**

``` php
use \Backpack\Base\BaseServiceProvider as Backpack;

public function setUp()
{
    Backpack::enqueueAsset('css', 'blockui', '/css/blockui.css');

    Backpack::enqueueAsset('js', 'jquery-ui', '/js/jquery-ui.js', ['jquery']);
    Backpack::enqueueAsset('js', 'angular-sortable', '/js/ng-sortable.js', ['angular']);
    Backpack::enqueueAsset('js', 'angular', '/js/angular.js', ['jquery']);
    Backpack::enqueueAsset('js', 'blockui', '/js/blockui.js',['jquery']);
}
```

I've added the functionality needed for a project, which will also help developers who want to add non conflicting input types to their code, however I've implemented it as a Trait on the BaseServiceProvider - which I'm pretty sure is not the best way to do this, however I do not know how you're meant to do something like this in Laravel, so feedback on how to implement this feature correctly would be lovely :)

### DIFF
--- a/src/Traits/Enqueue.php
+++ b/src/Traits/Enqueue.php
@@ -5,11 +5,11 @@ namespace Backpack\Base\Traits;
 trait Enqueue
 {
     public static $enqueuedScripts = [];
-    public static $enqueuedStyles  = [];
+    public static $enqueuedStyles = [];
 
     public static function enqueueAsset($type, $id, $source, $dependencies = [])
     {
-        switch($type){
+        switch ($type) {
             case 'css':
             case 'style':
                 return self::enqueueStyle($id, $source, $dependencies);
@@ -24,19 +24,18 @@ trait Enqueue
 
     public static function enqueueScript($id, $source, $dependencies)
     {
-        if(!str_contains($source, "//")){
+        if (!str_contains($source, '//')) {
             $path = '/'.ltrim($source, '/');
-            $v = filemtime(realpath(ltrim($path, "/")));
+            $v = filemtime(realpath(ltrim($path, '/')));
             $sourceComputed = asset($path.'?v='.$v);
-        }
-        else {
+        } else {
             $sourceComputed = $source;
         }
 
         $sourceObj = (object) [
             'id'     => $id,
             'source' => $sourceComputed,
-            'deps'   => $dependencies
+            'deps'   => $dependencies,
         ];
 
         self::$enqueuedScripts[$id] = $sourceObj;
@@ -46,26 +45,24 @@ trait Enqueue
 
     public static function enqueueStyle($id, $source, $dependencies)
     {
-        if(!str_contains($source, "//")){
+        if (!str_contains($source, '//')) {
             $path = '/'.ltrim($source, '/');
-            $v = filemtime(realpath(ltrim($path, "/")));
+            $v = filemtime(realpath(ltrim($path, '/')));
             $sourceComputed = asset($path.'?v='.$v);
-        }
-        else {
+        } else {
             $sourceComputed = $source;
         }
 
         $sourceObj = (object) [
             'id'     => $id,
             'source' => $sourceComputed,
-            'deps'   => $dependencies
+            'deps'   => $dependencies,
         ];
 
         self::$enqueuedStyles[$id] = $sourceObj;
 
         return $sourceObj;
     }
-
 
     private static function handleDependencies($items)
     {
@@ -74,13 +71,13 @@ trait Enqueue
         $dependenciesFound = [];
         $circularDependencyCounter = 0;
 
-        while(count($items) > count($resolutions) && $circularDependencyCounter < 20) {
+        while (count($items) > count($resolutions) && $circularDependencyCounter < 20) {
             $circularDependencyCounter++;
 
-            foreach($items as $itemIndex => $item) {
+            foreach ($items as $itemIndex => $item) {
 
                 //If I'm an existing depdendency - skip me!
-                if(isset($dependenciesFound[$item->id])) {
+                if (isset($dependenciesFound[$item->id])) {
                     continue;
                 }
 
@@ -88,16 +85,16 @@ trait Enqueue
                 $resolved = true;
 
                 //Check through each depdendency to see if its alrady desolved
-                if(isset($item->deps) && !empty($item->deps)) {
-                    foreach($item->deps as $dep) {
-                        if(!isset($dependenciesFound[$dep])) {
+                if (isset($item->deps) && !empty($item->deps)) {
+                    foreach ($item->deps as $dep) {
+                        if (!isset($dependenciesFound[$dep])) {
                             $resolved = false;
                             break;
                         }
                     }
                 }
 
-                if($resolved) {
+                if ($resolved) {
                     $dependenciesFound[$item->id] = true;
                     $resolutions[] = $item;
                 }
--- a/src/Traits/Enqueue.php
+++ b/src/Traits/Enqueue.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace Backpack\Base\Traits;
+
+trait Enqueue
+{
+    public static $enqueuedScripts = [];
+    public static $enqueuedStyles  = [];
+
+    public static function enqueueAsset($type, $id, $source, $dependencies = [])
+    {
+        switch($type){
+            case 'css':
+            case 'style':
+                return self::enqueueStyle($id, $source, $dependencies);
+                break;
+            case 'script':
+            case 'javascript':
+            case 'js':
+                return self::enqueueScript($id, $source, $dependencies);
+                break;
+        }
+    }
+
+    public static function enqueueScript($id, $source, $dependencies)
+    {
+        if(!str_contains($source, "//")){
+            $path = '/'.ltrim($source, '/');
+            $v = filemtime(realpath(ltrim($path, "/")));
+            $sourceComputed = asset($path.'?v='.$v);
+        }
+        else {
+            $sourceComputed = $source;
+        }
+
+        $sourceObj = (object) [
+            'id'     => $id,
+            'source' => $sourceComputed,
+            'deps'   => $dependencies
+        ];
+
+        self::$enqueuedScripts[$id] = $sourceObj;
+
+        return $sourceObj;
+    }
+
+    public static function enqueueStyle($id, $source, $dependencies)
+    {
+        if(!str_contains($source, "//")){
+            $path = '/'.ltrim($source, '/');
+            $v = filemtime(realpath(ltrim($path, "/")));
+            $sourceComputed = asset($path.'?v='.$v);
+        }
+        else {
+            $sourceComputed = $source;
+        }
+
+        $sourceObj = (object) [
+            'id'     => $id,
+            'source' => $sourceComputed,
+            'deps'   => $dependencies
+        ];
+
+        self::$enqueuedStyles[$id] = $sourceObj;
+
+        return $sourceObj;
+    }
+
+
+    private static function handleDependencies($items)
+    {
+        $items = array_unique($items, SORT_REGULAR); //Strip out duplicate requests
+        $resolutions = [];
+        $dependenciesFound = [];
+        $circularDependencyCounter = 0;
+
+        while(count($items) > count($resolutions) && $circularDependencyCounter < 20) {
+            $circularDependencyCounter++;
+
+            foreach($items as $itemIndex => $item) {
+
+                //If I'm an existing depdendency - skip me!
+                if(isset($dependenciesFound[$item->id])) {
+                    continue;
+                }
+
+                //We assume its found, unless stated below
+                $resolved = true;
+
+                //Check through each depdendency to see if its alrady desolved
+                if(isset($item->deps) && !empty($item->deps)) {
+                    foreach($item->deps as $dep) {
+                        if(!isset($dependenciesFound[$dep])) {
+                            $resolved = false;
+                            break;
+                        }
+                    }
+                }
+
+                if($resolved) {
+                    $dependenciesFound[$item->id] = true;
+                    $resolutions[] = $item;
+                }
+            }
+        }
+
+        return $resolutions;
+    }
+
+    public static function getScripts()
+    {
+        return self::handleDependencies(self::$enqueuedScripts);
+    }
+
+    public static function getStyles()
+    {
+        return self::handleDependencies(self::$enqueuedStyles);
+    }
+}
--- a/src/resources/views/layout.blade.php
+++ b/src/resources/views/layout.blade.php
@@ -30,6 +30,11 @@
     <!-- BackPack Base CSS -->
     <link rel="stylesheet" href="{{ asset('vendor/backpack/backpack.base.css') }}">
 
+    <!-- Enqueued CSS -->
+    @foreach( \Backpack\Base\BaseServiceProvider::getStyles() as $style )
+    <link rel="stylesheet" href="{{ $style->source }}" data-source-id="{{$style->id}}" data-deps="{{implode(',', $style->deps)}}">
+    @endforeach
+
     @yield('after_styles')
 
     <!-- HTML5 Shim and Respond.js IE8 support of HTML5 elements and media queries -->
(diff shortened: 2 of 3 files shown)


tabacitu:
Hmm... I'm not sure this is the best approach - to load assets from the controller. After all, the assets will be needed by something in a view, correct? So then, wouldn't it make more sense to "_add assets to the queue_" from that particular view? This would:
- keep both the asset and its usage in a single file;
- keep the controller asset-free, only managing the logic of the application;

If so, an alternative to "_add assets to queue_" already exists. In views that extend ```layout.blade.php```, you can add assets in the sections defined in layout.blade.php (```before_styles```,```after_styles```,```before_scripts```,```after_scripts```):
```
@section('after_styles')
  <link rel="stylesheet" href="{{ asset('vendor/adminlte/') }}/bootstrap/css/bootstrap.min.css">
@endsection
```
Of course, it doesn't test for existing views, but it's a very Laravel-way to do it.

What do you think, does this apply to your use case?

OwenMelbz:
Oh sure, they should not be defined in the controller, that was just an example to show it working I guess.

The way it would typically get used (again based off wordpress) would be when it's required, so lets say, for example the new Table field, I needed to load 3 external javascript libraries, 1 of them was angularjs, this happens in the `fields/table.blade.php`

I then want to use another field - which also needs angularjs, so the `fields/field.blade.php` gets included, because this ALSO has a call to angularjs, it will load it twice. Meaning my Table field will now break, because the angularjs instance has been overwritten.

So instead of doing

_-- fields/table.blade.php
```php
@push('crud_fields_scripts')
        {{-- YOUR JS HERE --}}
        <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.5.8/angular.min.js"></script>
        <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js"></script>
        <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/angular-ui-sortable/0.14.3/sortable.min.js"></script>
```

_-- fields/another-field.blade.php
```php
@push('crud_fields_scripts')
        {{-- YOUR JS HERE --}}
        <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js"></script>
```

you would instead do (ideally)

```php
@push('crud_fields_scripts')
        {{-- YOUR JS HERE --}}
       crud::enqueueScript('https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.5.8/angular.min.js', 'angularjs')
       crud::enqueueScript('https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js', 'jquery-ui')
       crud::enqueueScript('https://cdnjs.cloudflare.com/ajax/libs/angular-ui-sortable/0.14.3/sortable.min.js', 'angularjs-sortable')
```

_-- fields/another-field.blade.php
```php
@push('crud_fields_scripts')
        {{-- YOUR JS HERE --}}
       crud::enqueueScript('https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js', 'jquery-ui')
```

So when the layout.blade.php is run, it does collision detection and dependency checks, then displays the resolved scripts/styles from a loop similar to

```php
@foreach( crud::getScripts() as $script )
    <script src="{{ $script->source }}" data-source-id="{{$script->id}}" data-deps="{{implode(',', $script->deps)}}"></script>
@endforeach
```

tabacitu:
Ooooh. I get it now. Ok, I understand now, it makes sense to have that. 

I'll think about the syntax and do a small research, someone must have created a package for this. Even if it's not a great package, maybe we can learn from it.

I'll be back. Cheers!

tabacitu:
Found this: https://github.com/Stolz/Assets
Not saying it's any good, just didn't want to forget about it.

lloy0076:
@tabacitu @OwenMelbz I'm going to play devil's advocate here:

What is there was a command like:

```
const { mix } = require('laravel-mix');

mix.backpack(['base', 'crud', 'foo-bar-jiggy-module']);
```

...that got all the various backpack assets? It would work, at least, for the JS but I'm sure we could figure out the CSS too...

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
+<?php
+
+namespace Backpack\Base\Traits;
+
+trait Enqueue
+{
+    public static $enqueuedScripts = [];
+    public static $enqueuedStyles = [];
+
+    public static function enqueueAsset($type, $id, $source, $dependencies = [])
```
lloy0076: PHP Doc?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+                break;
+            case 'script':
+            case 'javascript':
+            case 'js':
+                return self::enqueueScript($id, $source, $dependencies);
+                break;
+        }
+    }
+
+    public static function enqueueScript($id, $source, $dependencies)
```
lloy0076: PHP Doc?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+    {
```
lloy0076: Should we check the inputs to make sure they're sensible here?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+        if (!str_contains($source, '//')) {
+            $path = '/'.ltrim($source, '/');
+            $v = filemtime(realpath(ltrim($path, '/')));
+            $sourceComputed = asset($path.'?v='.$v);
+        } else {
+            $sourceComputed = $source;
+        }
+
+        $sourceObj = (object) [
```
lloy0076: Is there any reason to do the double assignment here - it seems just as obvious to do:

```
self::$enqueuedScripts[$id] = (object) [ # ...
```

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+            'source' => $sourceComputed,
+            'deps'   => $dependencies,
+        ];
+
+        self::$enqueuedScripts[$id] = $sourceObj;
+
+        return $sourceObj;
+    }
+
+    public static function enqueueStyle($id, $source, $dependencies)
```
lloy0076: PHP Doc?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+    {
```
lloy0076: Should we validate the inputs?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+        if (!str_contains($source, '//')) {
```
lloy0076: The body of this method seems to be surprisingly similar to the method above? DRY?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+            $path = '/'.ltrim($source, '/');
+            $v = filemtime(realpath(ltrim($path, '/')));
+            $sourceComputed = asset($path.'?v='.$v);
+        } else {
+            $sourceComputed = $source;
+        }
+
+        $sourceObj = (object) [
```
lloy0076: See comments above.

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+            'source' => $sourceComputed,
+            'deps'   => $dependencies,
+        ];
+
+        self::$enqueuedStyles[$id] = $sourceObj;
+
+        return $sourceObj;
+    }
+
+    private static function handleDependencies($items)
```
lloy0076: PHP Doc?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+    {
+        $items = array_unique($items, SORT_REGULAR); //Strip out duplicate requests
```
lloy0076: What if it's NOT an array? i.e. should we validate that $items? Or at least give it a default value?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+        $resolutions = [];
+        $dependenciesFound = [];
+        $circularDependencyCounter = 0;
+
+        while (count($items) > count($resolutions) && $circularDependencyCounter < 20) {
```
lloy0076: Why 20 (as opposed to any other number)?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+            $circularDependencyCounter++;
+
+            foreach ($items as $itemIndex => $item) {
+
+                //If I'm an existing depdendency - skip me!
+                if (isset($dependenciesFound[$item->id])) {
+                    continue;
+                }
+
+                //We assume its found, unless stated below
```
lloy0076: it's (not its [sic]).

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+                $resolved = true;
+
+                //Check through each depdendency to see if its alrady desolved
```
lloy0076: it's

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+                if (isset($item->deps) && !empty($item->deps)) {
```
lloy0076: An `is_array` wouldn't hurt here either?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+                    foreach ($item->deps as $dep) {
+                        if (!isset($dependenciesFound[$dep])) {
+                            $resolved = false;
+                            break;
+                        }
+                    }
+                }
+
+                if ($resolved) {
```
lloy0076: Works but the paranoid would write `$resolved === true)`.

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+                    $dependenciesFound[$item->id] = true;
+                    $resolutions[] = $item;
+                }
+            }
+        }
+
+        return $resolutions;
+    }
+
+    public static function getScripts()
```
lloy0076: PHP Doc?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+    {
+        return self::handleDependencies(self::$enqueuedScripts);
```
lloy0076: Is `static::` better here?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public static function getStyles()
```
lloy0076: PHP Doc?

src/Traits/Enqueue.php
```diff
@@ -0,0 +1,116 @@
...
+    {
+        return self::handleDependencies(self::$enqueuedStyles);
```
lloy0076: `static::` ??

src/resources/views/layout.blade.php
```diff
@@ -30,6 +30,11 @@
     <!-- BackPack Base CSS -->
     <link rel="stylesheet" href="{{ asset('vendor/backpack/backpack.base.css') }}">
 
+    <!-- Enqueued CSS -->
+    @foreach( \Backpack\Base\BaseServiceProvider::getStyles() as $style )
+    <link rel="stylesheet" href="{{ $style->source }}" data-source-id="{{$style->id}}" data-deps="{{implode(',', $style->deps)}}">
```
lloy0076: Indenting is out.

The spacing in the implode is odd.

src/resources/views/layout.blade.php
```diff
@@ -117,21 +122,25 @@
 
         // Ajax calls should always have the CSRF token attached to them, otherwise they won't work
         $.ajaxSetup({
-                headers: {
-                    'X-CSRF-TOKEN': $('meta[name="csrf-token"]').attr('content')
-                }
-            });
+            headers: {
```
lloy0076: Is this a tabs vs spaces change?

src/resources/views/layout.blade.php
```diff
@@ -117,21 +122,25 @@
...
-          }
+            if ($(this).attr('href').startsWith(current_url) || current_url.startsWith($(this).attr('href'))) {
+                $(this).parents('li').addClass('active');
+            }
         });
     </script>
 
+    <!-- Enqueued Scripts -->
+    @foreach( \Backpack\Base\BaseServiceProvider::getScripts() as $script )
+    <script src="{{ $script->source }}" data-source-id="{{$script->id}}" data-deps="{{implode(',', $script->deps)}}"></script>
```
lloy0076: indenting, implode has issues with argument spaces.

AbbyJanke:
Okay. so I like this idea, we have a way to do it with fields currently (apart of CRUD) but I found a relatively easy way to check if we already included a specific stylesheet or javascript.

https://gist.github.com/AbbyJanke/bf64f5a62056b52bb94e9b9c379cb079

Usage Examples:

```js
<script type="text/javascript">
    if(!checkScripts('jquery.min.js')) {
        jQuery.getScript('YOURSCRIPT', function() {});
    }
</script>
```

```js
<script type="text/javascript">
    if(!checkStyles('backpack.base.css')) {
        $('head').append( $('<link rel="stylesheet" type="text/css" />').attr('href', 'YOURURL') );
    }
</script>
```

tabacitu:
Related issue in CRUD: https://github.com/Laravel-Backpack/CRUD/issues/1887
Package that does this: https://github.com/KosmosX/frontend

Personally I think Owen's implementation is cleaner. But I do think it should be a separate class. Maybe even a separate package, since it could be used on non-Backpack projects too.

I personally would see its features being:
- prevents loading one asset twice
- can help switch from CDN to local assets (aka CDN with fallback)
- can be called anywhere inside service provider, controller, view

Something like this:
```diff
-    Backpack::enqueueAsset('css', 'blockui', '/css/blockui.css');
+   Pipeline::add('css', 'blockui', '/css/blockui.css');

-    Backpack::enqueueAsset('js', 'jquery-ui', '/js/jquery-ui.js', ['jquery']);
+    Pipeline::add('js', 'jquery-ui', '/js/jquery-ui.js', ['jquery']);
```

```php

// Pipeline::add($stack, $id, $path, $dependencies);

Pipeline::add('css', 'blockui', '/css/blockui.css');
Pipeline::add('js', 'jquery-ui', '/js/jquery-ui.js', ['jquery']);

Pipeline::remove('css', 'blockui');
Pipeline::remove('js', 'jquery-ui');

// then we'd be able to do
Pipeline::get('css');
Pipeline::get('js');

// I don't see the pipelines as something set in stone, I think people should be able to create any number of pipelines
Pipeline::get('after_scripts');
Pipeline::get('after_styles');
Pipeline::get('metas');

// We could provide easy ways to output these pipelines, but I don't think that's necessary
// NOT providing outputs would actually be clearer, and would allow for endless customization
// for example, people could pass something else than strings to the add() command
// like an array, or an object

```

tabacitu:
Ooor we could just use this: https://github.com/orchestral/asset

OwenMelbz:
Go for it, 1 less thing to maintain

pxpm:
We should not reinvent the wheel. 
All the features you asked for are there. And seems like a very solid package. +1 :)

tabacitu:
I'm going to close this as it's so _so_ old that it will be impossible to merge. It's a pity it's 2020 and we still don't have a general-purpose system to enqueue assets in Laravel. 2016-2020 is a LONG time to go without one. Fortunately in the next major version of Backpack we'll have one, and we'll probably release it as a separate package too, so that people can use it outside Backpack too.

Thanks for the hard work on this @OwenMelbz - and @lloy0076 for the review . 

@OwenMelbz I know your preferred stack now includes NPM and React these days, but I'll get back to you in 6-8 months when we have a prototype for this (and for Backpack v12) - maybe I win you over from the dark side! 🤣 

Related:
- https://github.com/Laravel-Backpack/CRUD/issues/2283
- https://github.com/Laravel-Backpack/CRUD/pull/2652

PR #65, opened 2016-11-30
update font awesome version in layout.php (4.40->4.70)

zschuessler:
Font awesome version 4.4.0 to 4.7.0 (good number of missing icons between these versions)


tabacitu:
Thank you, merged!

PR #73, opened 2016-12-22
Refactored LoginController, fixed paths handling, logout redirect

pavoltanuska:
Well, this started as simple attempt to fix the logout procedure (Laravel redirects to '/' by default after logout), but once I was in I did the whole package.

- Reformated the docblocks a little bit
- Fixed the paths handling (whatever you had specified in the attributes was overwritten in the constructor, I've added some checks)
- Logout uses the default Laravel logout method, but redirects to location specified in LoginController

### DIFF
--- a/src/app/Http/Controllers/Auth/LoginController.php
+++ b/src/app/Http/Controllers/Auth/LoginController.php
@@ -30,7 +30,7 @@ class LoginController extends Controller
      * @var string
      */
     protected $loginPath = 'admin/login';
-    
+
     /**
      * Redirect here after successful login.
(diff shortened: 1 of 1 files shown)


tabacitu:
Very nice solution with the auth path overwriting, @pavoltanuska , thank you! 

I've merged it and implemented your solution for the other auth controllers, too. I think this will make it really easy for us to change the routes, not matter if we change the admin prefix or not. 

Thanks a lot, cheers!

pavoltanuska:
Good job, thank you too!

PR #88, opened 2017-01-08
[Enhancement] Added facility for an "edit profile" form - fixes #86

### LINKED ISSUE #86: User Profile - edit password functionality
Is there a plan to include user profile editing including edit password functionality in Base?

Like a cut down version of the User edit form in the PermissionManager but to allow the current user edit their own profile details.

OwenMelbz:
Based off https://github.com/Laravel-Backpack/Base/issues/86 here is a starting point for an edit profile function.

This adds a template, route and controller for an "Edit Account" functionality, by default it comes with the stock laravel auth fields. can be seen under `/admin/edit-profile`

It can be extended from the users application folder and the view modified/validation changed etc.

Heres simple screenshot of it

edit_account____backpack_admin

and the change password functionality

edit_account____backpack_admin 2


tabacitu:
Very nice! I'll slap an urgent label on it and merge it ASAP.

Cheers!

seandowney:
Hi, I'm just wondering what the status of this is? Thanks

tabacitu:
I think it's a much-needed functionality but I just haven't gotten to making the changes I want to it, so I can merge it:
1) I'd rename "EditAdminProfileController" to "ProfileController";
2) I wouldn't keep the "Edit Password" functionality in a modal inside a form, because it might be confusing for some users - I would either:
   A) place the "password" and "confirm password" inputs in the same form with name and email, with the placeholder text saying "_only type if you want to change the password_"
   B) place the change password in a different form than "change account info", separated by tabs or something;
   (I would probably go with option A, sounds easier to implement and simpler to the user)

I like how the modal looks - it looks great, but I think the "form inside a form" approach is better UX in ajax forms, when everything's submitted automatically. On regular forms, I find it confusing. It makes me (the user) think. And you know what they say in UX - "[Don't make me think](https://www.amazon.com/Dont-Make-Me-Think-Usability/dp/0321344758)" :-)

OwenMelbz:
I would ask kindly to not pollute generic namespaces with things like `ProfileController` - maybe something like `AdminProfileController`

The reason I actually went for the modal route was based off research we'd done, as well as developer flexibility.

I'll briefly explain why I think it should be either in its own self contained modal / page / form

The reason its pulled out of the rest of the form is to separate out logic and complexities.

By including them in the same form - you introduce extra logic and switches needed for developers - checking if theres data in the password fields, then complete extra work when it is detected, you end up having 1 controller endpoint managing 2 different things, which makes it harder to maintain, by having it separated it makes it more extendable.

By extracting the logic away from the main form you make it easier for users to focus on "the profile" rather than a "security" aspect of an account, until they need it.

You can see lots of examples where the password change is isolated such as

https://spark.laravel.com/settings#/security
spark

https://www.amazon.co.uk
amazon_co_uk_change_name__e-mail__password

https://www.paypal.com/myaccount/settings/password/edit/
paypal__settings

https://account.magento.com/customer/account/changepassword/
account_information

So. I think thats enough examples of why it should NOT be integrated into the same form.

Next up would be to consider tabs.

Tabs would seem sensible as you could have 2 different form elements on the page, each going to different endpoints.

It means you can keep the segregation for development, not worrying so much about the password logic, as well as keeping it in its own isolated view allowing users to focus simply on the password - and a chance for any password managers to hook in.

However, the biggest issue comes from the fact theres 2 forms on the page - if you make changes to your profile tab, then go to the password tab, and change the password, you will not save the profile fields. If you prompt the user to save the other information, then they lose the password changes.

So, that would leave only 2 remaining solutions.

* A Modal
* A Separate Page

I personally opted for the modal route, as it allows users to stay where they are, without navigating away from their profile edits, it also allows you to isolate the view when the modal is opened. (although you could remove the 2nd form, it doesn't really need it, was just there for semantics/accessibility/js support), and I felt it was more simple overall than having 2 pages.

So thats my reasoning, I wouldn't mind either of the last 2 solutions, however I think the modal is completely suitable for this :D

tabacitu:
**About the modal/tabs**
@OwenMelbz - very very comprehensive answer, thank you. You make a good point. Definitely convinced. So I think the best solution then would be to have a side-menu, that loads a different page? I'm thinking something like bootstrap pills?

screen_shot_2017-02-11_at_21_04_08

Bonus: this would also provide a place for the developer to introduce any other user-specific thing he adds.

Note to self: also include "current password" input, like all your examples seem to do, to prevent malicious usage.

**About the naming**
Ok, how about if it's included in the "Admin" folder In the application. It could be something like: ```app/Http/Controllers/Admin/UserProfileController.php```. It would still leave room for another one without admin, right (``````app/Http/Controllers/UserProfileController.php``````)?

OwenMelbz:
As you mentioned about separate pages, could we maybe...just use the pre-existing sidebar? Under the users avatar at the top left, we could have "My Account" then it expands to have a few?

Would seem less work :D

But yeah as long as its namespaced well enough not to cause conflict im all happy :D

tabacitu:
@OwenMelbz - I'm sure you'll be glad to know I'm _finally_ merging this :-)

I've done a lot of changes, though, please let me know what you think. If you could also take a look at the code and test, that would be great. @lloy0076 , you too, of course. Anybody else too, of course.

Changes:
- moved the "change password" modal into a separate form, like discussed above
- added "old password" input to the change password form (so if someone forgets his account logged in, at least he won't be locked out of his account);
- renamed the feature into "My Account" instead of "Profile", since we don't _actually_ have a user profile page;

Here are a few screenshots:
screen shot 2017-09-11 at 19 24 22
screen shot 2017-09-11 at 19 24 38
screen shot 2017-09-11 at 19 25 53

What do you guys think? Ready to merge into "update", unless you guys find something I missed.

Cheers!

OwenMelbz:
I'm not to sure about the photo in the top left - It might scare potential users off?

But apart from that seems all good from the screenshots! Will hopefully have a snoop deeper some time soon!

seandowney:
The screenshots and ideas look good to me. 

⁣Thank you

src/app/Http/Controllers/Auth/MyAccountController.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->data['user'] = Auth::user();
+
+        return view('backpack::auth.account.change_password', $this->data);
+    }
+
+    public function update(Request $request)
+    {
+        $user = Auth::user();
+
+        $validator = Validator::make($request->all(), [
```
OwenMelbz: I think maybe we should extract the validation logic away from the saving/processing of data - its likely users will want to add an extra field - it seems a shame for them to have to duplicate all this extra stuff?

src/app/Http/Controllers/Auth/MyAccountController.php
```diff
@@ -0,0 +1,88 @@
...
+            return redirect()->back()->withErrors($validator);
+        } else {
+            $user = Auth::user();
+
+            // check old password matches
+            if (!Hash::check($request->old_password, $user->password)) {
+                return redirect()->back()->withErrors(trans('backpack::base.old_password_wrong'));
+            }
+
+            $user->password = Hash::make($request->new_password);
```
OwenMelbz: maybe we should swap this hashing bit for the `bcrypt` helper - as this is most popular currently, and it has extra logic under the hood - so will be beneficial to use that and let the official laravel team handle it

src/app/Http/Controllers/Auth/MyAccountController.php
```diff
@@ -0,0 +1,88 @@
...
+
+            \Alert::success(trans('backpack::base.account_updated'))->flash();
+
+            return redirect()->back();
+        }
+    }
+
+    public function updatePassword(Request $request)
+    {
+        $validator = Validator::make($request->all(), [
```
OwenMelbz: Should we be enforcing the `required|min:6` as this might not be anywhere else? especially for the old password, we shouldnt enforce logic on that as it prevents cases where users had invalid passwords to update.

Also again maybe we should extract this logic out? I think if we just call smaller helper functions like `$this->getPasswordValidation()` this will allow users to just update the bits they want, rather than overloading the whole method?

src/app/Http/Controllers/Auth/MyAccountController.php
tabacitu: Domn. You're right, this isn't as clean as it could be. I've just split the validation into FormRequests, so it's squeaky-clean now :D

src/app/Http/Controllers/Auth/MyAccountController.php
tabacitu: Totally. I've removed the ```min:6``` and moved the validation to a FormRequest.

src/app/Http/Controllers/Auth/MyAccountController.php
tabacitu: Well ```Hash``` is just a Facade to that one. And I expect if the algorithm ever changes in Laravel, the facade will point to the new algorithm, which means this will still not need to be updated if we use the Facade.

tabacitu:
Thanks for the review and feedback, @OwenMelbz and @seandowney ! 

Merging into ```upgrade```. Cheers!

PR #106, opened 2017-02-15
pl/auth.php

djmassive:
Polish translation

### DIFF
--- a/src/resources/lang/pl/auth.php
+++ b/src/resources/lang/pl/auth.php
@@ -1,0 +1,21 @@
+<?php
+
+return [
+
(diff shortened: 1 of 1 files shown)


tabacitu:
Hi @djmassive ,

Unfortunately, this is the translation of a file that's not Backpack-specific, but Laravel-specific, so I can't merge it. The translation keys would be available under the "backpack" namespace, so they wouldn't be a drag-and-drop replacement for the langfile in Laravel. 

Best to include it in your own project. 

Looking forward to other PRs from you. Cheers!

PR #107, opened 2017-02-15
pl/passwords.php

djmassive:
Polish translation


OwenMelbz:
Hey, nice PR - but could you keep it in the same format as all the other languages? Would be weird to have polish in 1 format, and every other language in another 😄

djmassive:
Ok :-) i don't think it about like this :-)

W każdej chwili możesz wypisać się z subskrypcji - w tym celu wypisz ten adres e-mail z newslettera

tabacitu:
Hi @djmassive ,

Unfortunately, this is the translation of a file that's not Backpack-specific, but Laravel-specific, so I can't merge it. The translation keys would be available under the "backpack" namespace, so they wouldn't be a drag-and-drop replacement for the langfile in Laravel. 

Best to include it in your own project. 

Looking forward to other PRs from you. Cheers!

PR #115, opened 2017-03-15
Update layout.blade.php active and sidebar status

### LINKED ISSUE #104: Similar route name will make menu items <li> class "active" at the same time
If the route name is similar (../service & ../service-category), the side bar menu `<li>` are added `"active"` class together for "Service" and "Service Category" Sidebar menu.

``` javascript
// Set active state on menu element
var current_url = "{{ url(Route::current()->getUri()) }}";
$("ul.sidebar-menu li a").each(function() {
    if ($(this).attr('href').startsWith(current_url) || current_url.startsWith($(this).attr('href'))) {
        $(this).parents('li').addClass('active');
    }
});
```

The above code in the `layout.blade.php` need to be reviewed.

when the current_url is "`../../service`" or "`../../service-category`", the startWith function will give true when "Service" / “Service Category” menu is clicked.

MarcosBL:
Fixes #104 & also remembers sidebar collapsed state

### DIFF
--- a/src/resources/views/layout.blade.php
+++ b/src/resources/views/layout.blade.php
@@ -141,7 +141,7 @@
             });
 
         // Set active state on menu element
-        var current_url = "{{ Request::url() }}";
+        var current_url = "{{ Request::fullUrl() }}";
         $("ul.sidebar-menu li a").each(function() {
 			if (
 			   ($(this).attr('href').startsWith(current_url) || current_url.startsWith($(this).attr('href')))
--- a/src/resources/views/layout.blade.php
+++ b/src/resources/views/layout.blade.php
@@ -40,6 +40,15 @@
     <![endif]-->
 </head>
 <body class="hold-transition {{ config('backpack.base.skin') }} sidebar-mini">
+	<script type="text/javascript">
+		/* Recover sidebar state */
+		(function () {
+			if (Boolean(sessionStorage.getItem('sidebar-toggle-collapsed'))) {
+				var body = document.getElementsByTagName('body')[0];
+				body.className = body.className + ' sidebar-collapse';
+			}
+		})();
+	</script>
     <!-- Site wrapper -->
     <div class="wrapper">
 
@@ -112,6 +121,15 @@
 
     <!-- page script -->
     <script type="text/javascript">
+        /* Store sidebar state */
+        $('.sidebar-toggle').click(function(event) {
+          event.preventDefault();
+          if (Boolean(sessionStorage.getItem('sidebar-toggle-collapsed'))) {
+            sessionStorage.setItem('sidebar-toggle-collapsed', '');
+          } else {
+            sessionStorage.setItem('sidebar-toggle-collapsed', '1');
+          }
+        });
         // To make Pace works on Ajax calls
         $(document).ajaxStart(function() { Pace.restart(); });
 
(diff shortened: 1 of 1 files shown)


tabacitu:
Hi @MarcosBL ,

Totally agree with storing the sidebar state in the session. Should make for a better UX. @OwenMelbz  - do you feel the same?

I think this may be a good opportunity to move all JS in ```layout.blade.php``` that isn't PHP-dependant into an external JS file. Clean up the blade template. The problem is, we've used two standards:
- Backpack\Base uses ```backpack.base.css```, so a ```backpack.base.js``` would make sense;
- Backpack\CRUD uses ```crud\js\operation.js```;

I say we make this first JS file consistent with the Base package, and think about making it consistent across packages for the next breaking change (or v4, even). Do you guys agree?

PS. Sorry for checking in with you guys so much. Please tell me if it's annoying and I should stop. It's just been getting more and more difficult to think of all use cases myself to prevent untestable bugs. And two heads are usually better than one :-)

OwenMelbz:
Sure session sounds useful :)

Although I'm not sure why renaming the file is a breaking change? If people have it published it will use the old one, if they have not it will use the new one??

MarcosBL:
First of all, no worries, ask as much as you want, I had much fun and learnt a lot from your work: but take into account that I'm new to backpack, new to Laravel and new to programming good pratices in the first place :sob: 

This patch uses **sessionStorage** (browser local storage) so I can do the check without touching Laravel session nor needing an ajax call to save state; I think it's a "compromise" between usability & non-breaking changes. 

About the file naming, I just left the code in the layout.blade.php to make it easier to delete if someone doesn't use a collapsible sidebar, but agree both packages should use **external file** with the same path/name schema, and move as much as they can outside layout blade. 

It would be helpful as well to use some kind of "**assets queue**" processor ( Maybe Laravel Mix could do it even for custom fields https://laracasts.com/discuss/channels/general-discussion/elixir-combining-css-sheets ): AdminLTE dependencies + base dependencies + crud dependencies + custom columns/fields dependencies ends up being a bit of a nightmare to ensure you don't include assets twice, even 2 different versions of select2 for example (yes, happened to me!) etc... it would perfect to queue all of them both in crud & custom fields (with a blade directive to queue instead of echoing in the layout), and end with a single .js and a single .css

If the user need some more customizations (for example functions on the document.ready() event), we can always include a custom.js & custom.css that we queue in the last place prior to generating the final app.js & app.css

Just wondering!

pavoltanuska:
First of all, this looks very good, thanks @MarcosBL for the fix! 

I have one question though, how would you guys feel about changing the `var current_url = "{{ Request::url() }}";` to use `Request::fullUrl()` instead?

The reasoning is:
I regularly do administration of nested resources, for example Items and Categories, each Item belongs to a Category.
Administration of Categories is the usual stuff, for the Items I add a treeview to the sidebar, with "Items" being the parent and all the children are names of Categories and have `?category_id={id}` suffix in the href.

For illustration:
```
Sidebar
- Categories - /admin/category
- Items
-- All Categories - /admin/item
-- Category A - /admin/item?category_id=1
-- Category B - /admin/item?category_id=2
-- ...
```

In my CrudControllers I have:
```
if (request()->has('category_id') {
    $this->crud->addClause('where', 'category_id', request()->get('category_id');
}
```
which enables me to link to items belonging to a specific category right away.

(And I just realized, that I could probably use the Backpack's Filter functionality to get the most of it - well, you are now witnesses of a Rubber ducking session in real life).

When using only the `url()` method, the highlighted anchor is always "All Categories" (being the only one without GET parameters), but with `fullUrl()` everything works as expected.

The change seems like not breaking anything in the usual use cases and helps with this one.

Of course, I'm happy for any kind of feedback.

MarcosBL:
@pavoltanuska you are absolutely right, it's a non breaking change and covers a new case: already made that little change to this PR

tabacitu:
Hi @MarcosBL ,

Just merged this one. Sorry for taking so long, I've been away, than worked hard on client work AND on something you'll love when I reveal. Like love-love :-)

I'm now back to my usual schedule, so let's "Make Backpack great again!" (sad!).

Cheers!

MarcosBL:
Thank you ! And anxious about the love hype ! :D

PR #128, opened 2017-04-06
It's a ** 503 ** error - not a 500!

lloy0076:
Is there any reason why the 503.blade.php reports itself as a 500 [sic] error?

I'd propose to change it to a 503 seeing that's what it is :)


tabacitu:
NO! THIS IS ABSURD. Kidding, of course :-) Thanks for spotting this and submitting the PR. Merged, will be included in the next point update (early next week).

Cheers!

PR #129, opened 2017-04-06
replaced html comments with blade comments

alashow:
I'm gonna pull request to other repos (crud) too if this gets merged.

### DIFF
--- a/src/resources/views/auth/passwords/email.blade.php
+++ b/src/resources/views/auth/passwords/email.blade.php
@@ -1,6 +1,6 @@
 @extends('backpack::layout')
 
-<!-- Main Content -->
+{{-- Main Content --}}
 @section('content')
     <div class="row">
         <div class="col-md-8 col-md-offset-2">
--- a/src/resources/views/inc/menu.blade.php
+++ b/src/resources/views/inc/menu.blade.php
@@ -1,23 +1,23 @@
 <div class="navbar-custom-menu pull-left">
     <ul class="nav navbar-nav">
-        <!-- =================================================== -->
-        <!-- ========== Top menu items (ordered left) ========== -->
-        <!-- =================================================== -->
+        {{-- =================================================== --}}
+        {{-- ========== Top menu items (ordered left) ========== --}}
+        {{-- =================================================== --}}
 
-        <!-- <li><a href="{{ url('/') }}"><i class="fa fa-home"></i> <span>Home</span></a></li> -->
+        {{-- <li><a href="{{ url('/') }}"><i class="fa fa-home"></i> <span>Home</span></a></li> --}}
 
-        <!-- ========== End of top menu left items ========== -->
+        {{-- ========== End of top menu left items ========== --}}
     </ul>
 </div>
 
 
 <div class="navbar-custom-menu">
     <ul class="nav navbar-nav">
-      <!-- ========================================================= -->
-      <!-- ========== Top menu right items (ordered left) ========== -->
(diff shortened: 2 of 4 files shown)


tabacitu:
Hi @alashow ,

Hmm... I don't know about this. Why do you prefer Blade comment blocks over regular HTML comment blocks?

Cheers!

alashow:
Hey.

I prefer Blade comments because they are not printed to HTML output. Most of the comments are docs/developer related, and users don't need to see that. Also, regular HTML comment blocks makes HTML fatter :)

Cheers.

tabacitu:
Hmm... I honestly prefer to use HTML comments most of the time. I use them to track down which section I'm in (using Developer Tools or View Source). If I see an HTML comment, I can search for that HTML comment, whereas other HTML may be generated / have PHP / blade inside it.

I don't know. @OwenMelbz ? @pavoltanuska ? What's your opinion on this - HTML vs Blade comments?

OwenMelbz:
Maybe it depends on the comment :)

I would say that yeah they're developers only but if a normal user is snooping on source code, then so what if they see it lol

I also use HTML comments for debugging and seeing information, which is helpful

However if the comment is something like "todo- fix this later" then that should be a php comment.

However general helpful information that might be needed to see in the browser should be left in.

I think there's a place for both, and each comment should be addressed individually to see which is suitable.

pavoltanuska:
This one is tough and all comes down to personal/project preference to me. Personally, I usually don't care about which one is used, as long as it's used consistently.

But if I have to decide, I'm with @OwenMelbz's opininion - "each comment should be addressed individually to see which is suitable."

The rule of thumb for me is to comment the PHP/Blade parser functionality with Blade's comments and the HTML structure with HTML comments. E.g. if I want to state the reason why I'm looping some array exactly here and to explain why I'm doing it this way, I'm going to use the Blade comment.

As for the current arguments:

- I almost never look directly at the HTML output (if it's my code) - I'm using developer tools (don't we all) which handles all the navigation through the code without the need of the HTML comments.
- It's true, that the output is bigger ("fatter") using HTML comments, but does it really make a *substantial* difference? I don't know the exact numbers, but IMO the difference will be close to none on the usual pages (with lists with columns, forms with fields...).

If we go with the "consistency" there's another argument here - Backpack is built on top of the AdminLTE which is HTML+CSS template, not bound in any way to PHP (Laravel/Blade). Since it's plain HTML, there are only HTML comments used. So we've introduced the HTML comments in our files and we want to stay consistent. There's only one clear path from that - use HTML comments.

If we add some Blade functionality (`@if`, `@foreach`...) and want to comment their functionality specifically (not [the position they are on](https://github.com/Laravel-Backpack/Base/pull/129/files#diff-2ebc13c56479282b02cdc2e35b3f7837L76)) we should use Blade comments.

__TL;DR__
- use HTML comments for HTML stuff (position of elements, dividing sections, explanation of the use of HTML elements...)
- use Blade comments for Blade-specific functionality (conditional statements, loops...)

tabacitu:
@pavoltanuska  ..
giphy

I totally agree with you. I wanted to say _exactly_ the same thing, but was unable to put it into words. Very nice reply, and completely agree with the conclusion. As such, @alashow - I'm afraid I'm going to close this PR without merging it - these are all HTML comments, for HTML blocks, not for PHP/blade.

But please don't let this prevent you from submitting other PRs - they are very welcome and appreciated.

Cheers!

PR #134, opened 2017-04-21
Fix language files fallback problem

tabacitu:
## Please help me test this on Windows
I have tested this and it works on MacOS and Linux, but I doubt it will on Windows. **Can anybody please test and see if this also works on Windows?** How to test:
1. Require this Backpack\Base branch in your composer.json (dev-symlinks) and run ```composer update```.
2. Go through all Backpack packages in your vendor folder and comment out any translation loading you see (should look like this ```$this->loadTranslationsFrom(realpath(__DIR__.'/resources/lang'), 'backpack');```).
3. Delete or rename your ```resources/lang/vendor/backpack``` folder.
4. Go through your application and see if your language files are loaded. The most obvious would be in a  CRUD list view. If everything looks like crap ("backpack::crud.add" instead of "Add"), it doesn't work.

Thank you, thank you, thank you!

## Background story
The problem: 
- in all packages we load translation string with ```trans('backpack::base.something')```, ```trans('backpack::crud.something_else')```; that means there's only one namespace ("backpack") for all language strings;
- Laravel does not support loading files from multiple packages into a single namespace (I spent 3 separate days on this already, found zero ways to go around it); 
- so when we load the translation files from a second backpack package, we basically "forget" the translations from the first;
- we have gone around this limitation so far by publishing all language files, all the time; 
- this works, but because there's no fallback to the language files in the package, **every language file change is a breaking change**; this makes us wait for months before pushing something that should be a non-breaking change;

The solutions
1) Reorganize all language files and replace all strings, in all views. This would be a breaking change. And I really don't like the syntax. Last resort, in my opinion.
```diff
trans('backpack::crud.add');
trans('backpack-crud::crud.add');
```
2) [THIS PR] In Backpack\Base, create symlinks towards all language files, in all packages. Only load translations in Backpack\Base, no other packages. This will work no longer require us to publish all the language files. And if we do, there's a working fallback to all language files. So we can push updates to language files as much as we want. 

Thank you!


tabacitu:
@OwenMelbz @MarcosBL @pavoltanuska @lloy0076 - any of you guys have development environments on Windows?

pavoltanuska:
No Windows here, sorry.

But I have a suggestion, in case the Windows testing won't go well.

Instead of using symlinks we could use real files, looking like this:
```
// backpack/base/src/resources/lang/en/crud.php
<?php

if (file_exists(__DIR__ . '/../../../../../crud/src/resources/lang/en/crud.php')) {
    return require(__DIR__ . '/../../../../../crud/src/resources/lang/en/crud.php');
}

return [];
```

Is there any drawback I'm not able to see?

lloy0076:
I have windows but what am I meant to test? And my availability over the next 14 days could be terribly dreadful.

lloy0076:
Isn't there a Laravel helper to find the resources path?

tabacitu:
@pavoltanuska - I am ashamed to say that I did NOT consider that enough. I thought of it, tried it, then realised these php files would get published, and abandoned the solution. But it turns out we can't publish langfiles with symlinks either (league/flysystem doesn't support them), so your way is DEFINITELY better, as it's cross-platform.
 
@lloy0076 - no need to test on Windows anymore, I'll go with @pavoltanuska 's way and get back to you guys, on this same branch.

Thanks a lot for helping and keeping me in check guys, cheers!

MarcosBL:
Was coming to suggest the realpath() solution, as Windows support for symlinks is bad, being polite 😂

tabacitu:
It is DONE! Thanks a lot, guys. We can now easily merge PRs with language files :-)

Cheers!

PR #137, opened 2017-04-26
Design Facelift

tabacitu:
Made a few design improvements to make AdminLTE look less outdated (compared to other premium admin themes):
- bigger texts
- more space

Also:
- added Gravatar instead of placeholder image with letter;
- cleaned up sidebar so there are less elements;

This new design is introduced as an "overlay" (a new css file) I called "bold". Since these changes are very subjective, I expect other people to want to create their own overlays, and easily drop them in all projects to get their preferred look-and-feel.

What do you guys think? Better/worse than before?
screen shot 2017-04-26 at 12 45 43
screen shot 2017-04-26 at 12 45 10


erik-ropez:
Looks nice. Not sure about additional space above "Monsters".

victorlap:
I don't know if adding the gravatar laravel package as a dependency is the right thing to do. It would force the use and loading of an additional service provider for every request and all laravel projects who use backpack, even though likely most of the people have their own images they want to show in the sidebar, or maybe none at all.

I would consider removing gravatar at all, or maybe just use a link without the need of the service provider and additional assets and packages.

tabacitu:
Damn it, @victorlap , why do you have to be all rational?! :-) Sounds to me like you're right - not all projects will need it... damn it :-)

I'll be back in town in about 1 week - i'll give more thought to this until then. Curious what others think. 

Cheers! Thanks for keeping me in check!

lloy0076:
It looks all right to me :)

bloggervista:
I also don't want to load gravatar for every request. I hope this get merged without gravatar :)

eduardoarandah:
I suggest leaving gravatar as commented code, plus local avatar image as commented code. 

That corner feels like "this is me and my place in that system" 
So I would place a "my profile" link 

I don't know if that's out of scope for backpack, but every system needs a "my profile" to change email, name and password 

What do you think?

eduardoarandah:
How about font-awesome ?
http://imgur.com/a/qGw6n

Light, nice, easy to change color and size. May default to theme color, gray or gradient

`<i class="fa fa-user-circle-o" aria-hidden="true"></i>`

      .main-sidebar .user-panel .fa{
        color: #d8d8d8; font-size: 46px;margin-left: 5px;
      }
      .sidebar-collapse .main-sidebar .user-panel .fa{
        color: #d8d8d8;
        font-size: 20px;
        margin-left: 5px;
        margin-top: 5px;
      }

And leaving as a setting:
- font-awesome
- gravatar
- avatar field in user model

Intuition is, clicking in it, should send you to /admin/my-profile to upload avatar (or gravatar), but maybe that's outside backpack responsability

tabacitu:
Ok, I think this is finally ready to be merged :-)

I've taken your feedback into account and I've created a new option in ```config/backpack/base.php```:
```php
    // What kind of avatar will you like to show to the user?
    // Default: gravatar (automatically use the gravatar for his email)
    // Other options:
    // - placehold (generic image with his first letter)
    // - example_method_name (specify the method on the User model that returns the URL)
    'avatar_type' => 'gravatar',
```

I personally don't mind loading an image from gravatar on every pageload. Keep in mind we used to do the same thing with placehold.it :-) So I left "gravatar" as default, because I expect this to bring the best user experience - it's a best-practice in UI to show the user his picture if you can, so that he easily understands he's logged in as himself. Plus, it takes root in a deep bias - we like ourselves, we like our picture and we like our name. Subconsciously, anything that has our name and picture on it is instantly better :-) By 1% most likely, but better nonetheless :-) If we can take advantage of this bias and make our users happier - and like the admin panels we build, why not do it?

Bottom line, we now have three options:
- gravatar (default)
- placehold (the old one)
- custom method on the User model (it will call whatever method you want on the user model that returns a URL; this way you can easily customize the avatar to whatever image you want, if you don't like the options above)

I think this is a pretty simple way to keep everyone happy, wouldn't you agree? :-)

Cheers!

tabacitu:
A little offtopic: @eduardoarandah completely agree with you, next to the user image would be a great place to put a "preferences" link. Will do.

eduardoarandah:
Agree, 

Quick thought:

A myprofile option

    'avatar_type' => 'myprofile',

Paired with a Laravel-Backpack package "MyProfile"

That uses a CRUD controller similar to the one in package-manager, the only difference being it only loads your own user.

Instead of  this route
http://laravel.dev/app/admin/user/3/edit

something like 
http://laravel.dev/app/admin/myprofile/edit

This is permission manager when editing a user:

abc

It just needs an image field and 

aand-its-done

tabacitu:
Hmm... yes, should be careful about not letting the user change his permissions, or anything like that, though :-) So maybe a separate interface for editing his own user info and/or avatar then so no mistakes can be done there...

RedAB:
Hi, do you think that updating the AdminLTE dependency to the last 2.4 version would be a problem ? According [to the upgrade guide](https://adminlte.io/docs/2.4/upgrade-guide), there doesn't seem to be any breaking change. Thanks for answering !

PR #141, opened 2017-05-02
Add option to switch authentication guards

### LINKED ISSUE #122: user_model_fqn used in Register, but not in Login
Hi everyone,
I'm facing a serious problem with using a custom user model in Backpack.

I create a new model, migrated it based on basic laravel user model, and specified in config/backpack/base.php the new model name.

I found out this configuration is used in registration (A user is correctly created and I am correctly auto-logged after creation), but if I logout and try to login again it just says matching credentials were not found.

Tried many solutions(Creating guards, providers, controllers, even editing Backpack core), but I'm just getting nowhere.

Is there any help you could provide me? I need no guards, middlewares or whatever, I just need to use a different user model.

jorenvanhee:
This fixes #122.

I basically made it possible to switch authentication guards for backpack. This makes it possible to use a different guard besides the default one (from `config/auth.php`).

For instance, you have a guard for the users on the public facing part of the website. And you have a different guard for the admins of the website. So a `user` guard and an `admin` guard. If the `user` guard is the default one, then it was impossible to use the `admin` guard with backpack. This pull request makes it possible to set a specific guard for backpack.

There was already a config option `user_model_fqn` in backpack, but this only changed the behaviour of the user registration. More info about that in issue #122.

~~This should be a non-breaking change because the `user_model_fqn` config option is still present and being used. We could remove it and fetch the used model from the guard configuration, but that would be a breaking change.~~

This is a breaking change because the views are different now, and they have to be republished. Since it is a breaking change, I will also look into removing the `user_model_fqn` option.

Actually, the views only need to be republished, if you use the new `guard` option. The old views just use the default guard. So I'm not sure if this is breaking or not...

### DIFF
--- a/src/BaseServiceProvider.php
+++ b/src/BaseServiceProvider.php
@@ -3,6 +3,7 @@
 namespace Backpack\Base;
 
 use Illuminate\Routing\Router;
+use Illuminate\Support\Facades\View;
 use Illuminate\Support\ServiceProvider;
 use Route;
 
@@ -27,6 +28,8 @@ class BaseServiceProvider extends ServiceProvider
         $this->loadViewsFrom(resource_path('views/vendor/backpack/base'), 'backpack');
         // - then the stock views that come with the package, in case a published view might be missing
         $this->loadViewsFrom(realpath(__DIR__.'/resources/views'), 'backpack');
+
+        View::composer('backpack::*', \Backpack\Base\app\Http\ViewComposers\AuthComposer::class);
 
         $this->loadTranslationsFrom(realpath(__DIR__.'/resources/lang'), 'backpack');
 
--- a/src/app/Http/Controllers/Auth/ForgotPasswordController.php
+++ b/src/app/Http/Controllers/Auth/ForgotPasswordController.php
@@ -4,6 +4,7 @@ namespace Backpack\Base\app\Http\Controllers\Auth;
 
 use Backpack\Base\app\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\SendsPasswordResetEmails;
+use Illuminate\Support\Facades\Password;
 
 class ForgotPasswordController extends Controller
 {
@@ -29,7 +30,10 @@ class ForgotPasswordController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('guest');
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
+
+        $this->middleware("guest:$guard");
     }
 
     // -------------------------------------------------------
@@ -46,5 +50,13 @@ class ForgotPasswordController extends Controller
         $this->data['title'] = trans('backpack::base.reset_password'); // set the page title
 
         return view('backpack::auth.passwords.email', $this->data);
+    }
+
+    public function broker()
+    {
+        $passwords = config('backpack.base.passwords')
+            ?: config('auth.defaults.passwords');
+
+        return Password::broker($passwords);
     }
 }
--- a/src/app/Http/Controllers/Auth/LoginController.php
+++ b/src/app/Http/Controllers/Auth/LoginController.php
@@ -2,10 +2,10 @@
 
 namespace Backpack\Base\app\Http\Controllers\Auth;
 
+use Auth;
 use Backpack\Base\app\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\AuthenticatesUsers;
 use Illuminate\Http\Request;
-use Auth;
 
 class LoginController extends Controller
 {
--- a/src/app/Http/Controllers/Auth/LoginController.php
+++ b/src/app/Http/Controllers/Auth/LoginController.php
@@ -5,6 +5,7 @@ namespace Backpack\Base\app\Http\Controllers\Auth;
 use Backpack\Base\app\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\AuthenticatesUsers;
 use Illuminate\Http\Request;
+use Auth;
 
 class LoginController extends Controller
 {
@@ -31,7 +32,10 @@ class LoginController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('guest', ['except' => 'logout']);
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
+
+        $this->middleware("guest:$guard", ['except' => 'logout']);
 
         // ----------------------------------
         // Use the admin prefix in all routes
@@ -80,5 +84,13 @@ class LoginController extends Controller
 
         // And redirect to custom location
         return redirect($this->redirectAfterLogout);
+    }
+
+    protected function guard()
+    {
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
+
+        return Auth::guard($guard);
     }
 }
--- a/src/app/Http/Controllers/Auth/RegisterController.php
+++ b/src/app/Http/Controllers/Auth/RegisterController.php
@@ -2,11 +2,11 @@
 
 namespace Backpack\Base\app\Http\Controllers\Auth;
 
+use Auth;
 use Backpack\Base\app\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\RegistersUsers;
 use Illuminate\Http\Request;
 use Validator;
-use Auth;
 
 class RegisterController extends Controller
 {
--- a/src/app/Http/Controllers/Auth/RegisterController.php
+++ b/src/app/Http/Controllers/Auth/RegisterController.php
@@ -6,6 +6,7 @@ use Backpack\Base\app\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\RegistersUsers;
 use Illuminate\Http\Request;
 use Validator;
+use Auth;
 
 class RegisterController extends Controller
 {
@@ -30,7 +31,10 @@ class RegisterController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('guest');
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
+
+        $this->middleware("guest:$guard");
 
         // Where to redirect users after login / registration.
         $this->redirectTo = property_exists($this, 'redirectTo') ? $this->redirectTo
@@ -112,5 +116,13 @@ class RegisterController extends Controller
         $this->guard()->login($this->create($request->all()));
 
         return redirect($this->redirectPath());
+    }
+
+    protected function guard()
+    {
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
+
+        return Auth::guard($guard);
     }
 }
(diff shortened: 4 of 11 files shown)


soham2008xyz:
Have you seen https://github.com/Laravel-Backpack/Base/pull/87? Exactly similar to our use case here, hasn't been merged yet.

lloy0076:
@soham2008xyz @jorenvanhee - it seems that https://github.com/Laravel-Backpack/Base/pull/87 is similar but not quite the same although I do agree that they are _probably_ addressing the same issue.

 I think the main point of difference is that 87 includes the guard whereas this one doesn't.

src/app/Http/Controllers/Auth/ForgotPasswordController.php
```diff
@@ -29,7 +30,10 @@ class ForgotPasswordController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('guest');
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: Is there a reason this is on a new line?

src/app/Http/Controllers/Auth/ForgotPasswordController.php
```diff
@@ -47,4 +51,12 @@ public function showLinkRequestForm()
 
         return view('backpack::auth.passwords.email', $this->data);
     }
+
+    public function broker()
```
lloy0076: PHP doc?

src/app/Http/Controllers/Auth/ForgotPasswordController.php
```diff
@@ -47,4 +51,12 @@ public function showLinkRequestForm()
...
+    {
+        $passwords = config('backpack.base.passwords')
+            ?: config('auth.defaults.passwords');
```
lloy0076: NL?

src/app/Http/Controllers/Auth/LoginController.php
```diff
@@ -31,7 +32,10 @@ class LoginController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('guest', ['except' => 'logout']);
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: NL?

src/app/Http/Controllers/Auth/LoginController.php
```diff
@@ -81,4 +85,12 @@ public function logout(Request $request)
         // And redirect to custom location
         return redirect($this->redirectAfterLogout);
     }
+
+    protected function guard()
```
lloy0076: PHP Doc?

src/app/Http/Controllers/Auth/LoginController.php
```diff
@@ -81,4 +85,12 @@ public function logout(Request $request)
...
+    {
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: NL?

src/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -30,7 +31,10 @@ class RegisterController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('guest');
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: NL?

src/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -113,4 +117,12 @@ public function register(Request $request)
 
         return redirect($this->redirectPath());
     }
+
+    protected function guard()
```
lloy0076: PHP Doc?

src/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -113,4 +117,12 @@ public function register(Request $request)
...
+    {
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: NL?

src/app/Http/Controllers/Auth/ResetPasswordController.php
```diff
@@ -30,7 +32,10 @@ class ResetPasswordController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('guest');
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: NL?

src/app/Http/Controllers/Auth/ResetPasswordController.php
```diff
@@ -58,4 +63,20 @@ public function showResetForm(Request $request, $token = null)
             ['token' => $token, 'email' => $request->email]
         );
     }
+
+    public function broker()
```
lloy0076: PHP Doc?

src/app/Http/Controllers/Auth/ResetPasswordController.php
```diff
@@ -58,4 +63,20 @@ public function showResetForm(Request $request, $token = null)
...
+    {
+        $passwords = config('backpack.base.passwords')
+            ?: config('auth.defaults.passwords');
```
lloy0076: NL?

src/app/Http/Controllers/Auth/ResetPasswordController.php
```diff
@@ -58,4 +63,20 @@ public function showResetForm(Request $request, $token = null)
...
+
+        return Password::broker($passwords);
+    }
+
+    protected function guard()
```
lloy0076: PHP Doc?

src/app/Http/Controllers/Auth/ResetPasswordController.php
```diff
@@ -58,4 +63,20 @@ public function showResetForm(Request $request, $token = null)
...
+    {
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: NL?

src/app/Http/Middleware/Admin.php
```diff
@@ -12,12 +12,14 @@ class Admin
...
      * @param \Closure                 $next
-     * @param string|null              $guard
      *
      * @return mixed
      */
-    public function handle($request, Closure $next, $guard = null)
+    public function handle($request, Closure $next)
     {
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: NL?

src/app/Http/ViewComposers/AuthComposer.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Backpack\Base\app\Http\ViewComposers;
+
+use Illuminate\Support\Facades\Auth;
+use Illuminate\View\View;
+
+class AuthComposer
+{
+    public function compose(View $view)
```
lloy0076: PHP Doc?

src/app/Http/ViewComposers/AuthComposer.php
```diff
@@ -0,0 +1,19 @@
...
+    {
+        $guard = config('backpack.base.guard')
+            ?: config('auth.defaults.guard');
```
lloy0076: NL?

src/resources/views/inc/menu.blade.php
```diff
@@ -19,7 +19,7 @@
 
       <!-- <li><a href="{{ url('/') }}"><i class="fa fa-home"></i> <span>Home</span></a></li> -->
 
-        @if (Auth::guest())
+        @if ($backpackAuth->guest())
```
lloy0076: Where did `$backpackAuth` come from?

jorenvanhee: From the [AuthComposer](https://github.com/jorenvanhee/Base/blob/different-user-model-issue-fix/src/app/Http/ViewComposers/AuthComposer.php).

src/app/Http/Controllers/Auth/ResetPasswordController.php
jorenvanhee: Added the newline here to keep width of the line under 80 characters. I also did it in some other places where it may not be needed. I thought it was readable like that. But they can be removed of course :)

src/app/Http/Controllers/Auth/ResetPasswordController.php
jorenvanhee: Can add that for sure, if there's interest in merging this PR.

jorenvanhee:
Is there something more that I can do to get this PR merged?

If this one gets merged, I will also have to update the `authorize` method in the `CrudRequest` class from Backpack/CRUD.

OwenMelbz:
@jorenvanhee this looks like a duplicate of https://github.com/Laravel-Backpack/Base/pull/87 which has had some discussion over.

tabacitu:
@jorenvanhee - I'm sorry, I've been hacking away at client projects and preparing something super exciting for the Backpack community that will be revealed soon, so I haven't merged PRs for a while. But I'm back now!

@OwenMelbz - is this a complete replacement for #87? Cause if it is, I think I like it a bit better, and it would be a non-breaking change, so it's ready to test and merge. I especially like that:
- it avoids using a helper;
- the guard usage has better readability
- ```@if ($backpackAuth->guest())``` over ```@if (!backpack_admin())```, because it allows you to call ```->check()``` and other methods too;
- the configuration file seems easier to understand to me (just replace the Auth middleware if you need to);
- it falls back to configuration options from Laravel itself (```config/auth.php```);

Honestly, I think "_pretty awesome_" is the word for this PR :-) 
Really nice job @jorenvanhee :-)

src/BaseServiceProvider.php
```diff
@@ -28,6 +29,8 @@ public function boot(\Illuminate\Routing\Router $router)
         // - then the stock views that come with the package, in case a published view might be missing
         $this->loadViewsFrom(realpath(__DIR__.'/resources/views'), 'backpack');
 
+        View::composer('backpack::*', \Backpack\Base\app\Http\ViewComposers\AuthComposer::class);
```
tabacitu: @jorenvanhee - I'm wondering if this is broad enough. Wouldn't this make the ```$backpackAuth``` guard available only in views that are loaded within the backpack namespace?

If so, I see the following problem:
- developer uses $backpackAuth when he overwrites backpack views;
- developer create supplementary views, that he considers to be part of the admin panel; he tries to use $backpackAuth there too; he expects ```$backpackAuth``` to be there too, but it's not;

I may be mistaken, though. It's been known to happen :-)

tabacitu:
@jorenvanhee , @lloy0076 - I've merged this into a separate branch and done @lloy0076 's suggestions.

Let's move the conversation there, please.

tabacitu:
Closing this - let's move the conversation there.

OwenMelbz:
@tabacitu just this asking if its a replacement for the other PR - answer is no lol, it does not address the issues we face.

Regards to the comments of

- it avoids using a helper;
- the guard usage has better readability
- @if ($backpackAuth->guest()) over @if (!backpack_admin()), because it allows you to call ->check() and other methods too;
- the configuration file seems easier to understand to me (just replace the Auth middleware if you need to);
- it falls back to configuration options from Laravel itself (config/auth.php);

Responses being
- We've just introduced helpers - so that argument I assume is now void?
- Not sure this one is entirely true :D even adds another dependency of passing data to the view which means more files will need to be updated for the user
- The helper was designed to get the user instance - not query the guard - you can still do that
- Again the issue was not that we want a custom guard - the issue is separation of sessions - you cannot be logged into only 1 without customly writing this yourself which adds annoying overheads.
- The user can use the default guard if they want

Issue #148, opened 2017-05-25
Active sidebar menu item does not take query string into account


DelightedD0D:
Given the below snippet, from my `sidebar.blade.php`, if I click on any of the 4 menu items, when the page loads, all 4 menu items are set to `active`. 

This happens because they are links for the same route with different query parameters and query parameters are not considered in the current implementation.

```
                <li><a href="{{ url('/diagonals/docs') }}"><i class="fa fa-book "></i> <span>Docs</span></a></li>
                <li class="treeview">
                  <a href="#"><i class="fa fa-line-chart"></i> <span>Graphs</span> <i class="fa fa-angle-left pull-right"></i></a>
                  <ul class="treeview-menu">
                    <li><a href="{{ url('/diagonals/graph') }}"><i class="fa fa-clock-o "></i> <span>Hourly Graph </span></a></li>
                    <li><a href="{{ url('/diagonals/graph?freq=daily') }}"><i class="fa fa-calendar-o"></i> <span>Daily Graph </span></a></li>
                    <li><a href="{{ url('/diagonals/graph?cat=Marketing') }}"><i class="fa fa-bar-chart"></i> <span>Marketing Graph </span></a></li>
                    <li><a href="{{ url('/diagonals/graph?maponly=1') }}"><i class="fa fa-share-alt-square"></i> <span>Relationship Graph</span></a></li>
                  </ul>
                </li>
```

![image](https://cloud.githubusercontent.com/assets/8271955/26470564/c88d9a4a-4163-11e7-89d7-635dc7d2c83f.png)

tabacitu:
Hi @DelightedD0D ,

Just merged your PR, thank you, a ```composer update``` should do it for you, if you haven't overwritten the ```layout.blade.php``` file.

Cheers!

jrpub:
Hi all,

I think there is still a bug here. 

In layout.blade.php, the following line: 

```
var current_url = "{{ Request::fullUrl() }}";
```
should probably be:

```
var current_url = "{{ Request::url() }}";
```
**fullUrl** returns the full URL with parameters, contrary to **url()**. 

With the line `full_url = current_url+location.search;` the parameters will be added twice according to my tests.

DelightedD0D:
@jrpub Yeah, if you look at the actual commit, it originally was just url(), but was later updated to fullUrl() when other functionality was added in https://github.com/Laravel-Backpack/Base/pull/115 and we didnt catch the conflict. 

[...]

PR #149, opened 2017-05-25
Support query string when setting active menu item

DelightedD0D:
Addresses #148


tabacitu:
Thank you, @DelightedD0D . Merged it!

jrpub:
Hi all,

I think there is still a bug here. 

In layout.blade.php, the following line: 

```
var current_url = "{{ Request::fullUrl() }}";
```
should probably be:

```
var current_url = "{{ Request::url() }}";
```
**fullUrl** returns the full URL with parameters, contrary to **url()**. 

With the line `full_url = current_url+location.search;` the parameters will be added twice according to my tests.

PR #151, opened 2017-05-30
Use Mix instead of Elixir for Laravel 5.4

bloggervista:
For Laravel 5.3  we still need to use Elixir for 5.4 comes with mix
Use Mix instead of Elixir for Laravel 5.4

### DIFF
--- a/src/resources/views/layout.blade.php
+++ b/src/resources/views/layout.blade.php
@@ -143,6 +143,6 @@
     @yield('after_scripts')
 
     <!-- JavaScripts -->
-    {{-- <script src="{{ elixir('js/app.js') }}"></script> --}}
+    {{-- <script src="{{ mix('js/app.js') }}"></script> --}}
 </body>
 </html>


src/resources/views/layout.blade.php
```diff
@@ -143,6 +143,6 @@
     @yield('after_scripts')
 
     <!-- JavaScripts -->
-    {{-- <script src="{{ elixir('js/app.js') }}"></script> --}}
+    {{-- <script src="{{ mix('js/app.js') }}"></script> --}}
```
lloy0076: If this is commented out, why not just **remove** this entirely?

lloy0076:
The script is commented out in Blade - why's it there in the first place?

OwenMelbz:
Guess was commented out from a core copy/paste and was never needed! can probably drop it competely

tabacitu:
As far as I remember, the talk back then was wether to use Mix/Elixir/Webpack by default or just make it easy for people to use it if they want. And the conclusion was to have this, but commented out. If you want it, uncomment it. Otherwise just leave it like that.

I believe Laravel does the same thing in its default generated blade files, so I say let's keep it commented. If you agree, I'll merge it in the "upgrade" branch, where we'll only provide L5.5 support.

bloggervista:
@tabacitu Yes  I guess we should either keep it with mix or remove it. Elixir keyword will be unknown to new devs and laravel 5.5 LTS by default uses mix.

PR #152, opened 2017-05-30
Use Stacks instead of yield

bloggervista:
It is better for future than yield as in stack isn't it?

### DIFF
--- a/src/resources/views/layout.blade.php
+++ b/src/resources/views/layout.blade.php
@@ -12,7 +12,7 @@
       {{ isset($title) ? $title.' :: '.config('backpack.base.project_name').' Admin' : config('backpack.base.project_name').' Admin' }}
     </title>
 
-    @yield('before_styles')
+    @stack('before_styles')
 
     <meta content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no" name="viewport">
     <!-- Bootstrap 3.3.5 -->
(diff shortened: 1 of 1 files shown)


lloy0076:
@bloggervista 

1. I find stacks more intuitive
2. Stacks are better than *yield* in the sense that one can *push* to them multiple times from different view templates

@tabacitu - do you have an opinion?

bloggervista:
@lloy0076 
Your Point no 2 was exactly my thought :)

bloggervista:
@lloy0076  please don't merge this immediately merge it in next version. Because it may seem it will not change anything but it will affect backpack crud because i found backpack crud used @section('before_syles') like that. Before merging we need to make it backpack crud friendly right?

lloy0076:
@bloggervista - AFAICT one can use `@stack`s and `@sections` but I'd be incline to suggest that the *before_styles* section be deprecated.

bloggervista:
@lloy0076  I also want them to be deprecated but i don't know this before_styles will ripple to other package like only crud or other packages only. I think we need to update that first. But it seems the author of this package is busy :(

tabacitu:
@bloggervista , @lloy0076 - totally agree with you guys on the _stack vs yield_ debate. The only reason we've used yield so far is because this project has started before "stacks" existed.

That being said, I definitely want to merge this, but it _does_ represent a breaking change, so we need to time this just right. We made a point in bundling all breaking changes together, and launching them once Laravel itself launches another version. That way, Backpack developers only need to sit down and upgrade once: first Laravel, then Backpack.

So it should be another about 2 months until this gets merged and made public, sorry.

Cheers!

bloggervista:
@tabacitu  Can we add latest feature in another branch?

tabacitu:
@bloggervista , @lloy0076 - I think this was a great change. And now that 0.9 is almost ready for launch, I think it's a really good moment to introduce it. I haven't been able to merge this as the source of the PR is gone, but I've made a separate PR here https://github.com/Laravel-Backpack/Base/pull/260 that has it, with a twist - it's now non-breaking.

Thanks a lot for this @bloggervista .
Cheers!

bloggervista:
Thanks tabacu for this package :)

Issue #131, opened 2017-04-09
Consistent extend-package command


tabacitu:
I think the next big improvement in the Backpack packages will be to make it easier to extend them, and let the developer modify all classes to his liking. Right now that's a little difficult, and we don't have documentation for each package on how to do that. This proposal aims to bring a consistent way to extend all Backpack packages.

Packages involved (and what will be extended):
- Base (DashboardController, Auth controllers, admin middleware, ResetPasswordNotification)
- CRUD (CrudController, CrudRequest, CrudPanel)
- LogManager (LogController)
- PageManager (PageCrudController, PageRequest, Models\Page)
- PermissionManager (controllers, requests, models)
- NewsCrud (controllers, requests, models)
- MenuCrud (controller, request, model)

Commands:
- php artisan backpack:extend-base
- php artisan backpack:extend-crud
- php artisan backpack:extend-logmanager
- php artisan backpack:extend-pagemanager
- php artisan backpack:extend-permissionmanager
- php artisan backpack:extend-newscrud
- php artisan backpack:extend-menucrud

Each command would:
- publish blank controllers in ```app/Http/Controllers/Admin/``` that extend the ones in the package;
- publish blank models in ```app/Models/``` that extend the ones in the package;
- publish blank requests in ```app/Requests/``` that extend the ones in the package;
- publish blank middleware in ```app/Middleware/``` that extend the ones in the package;
- publish blank php classes in ```app/Library/``` that extend the other classes in the packages (CrudPanel for example);
- create a routes file inside ```routes/backpack``` with the routes pointing to the new controllers;
- echo in the command line any more steps that the developer needs to do to finish extending the package (no need to look over the documentation to find out);

File naming:
- the extended files will have the exact same name as the ones they extend; the only difference will be the namespace (one is inside ```App/Http```, one is inside ```Backpack/PageManager/app/Http/etc```).

Result:
Basically running an extend command will publish all the blank classes you need to extend a package, and the routes file to make the application use those new files. For most packages, this single command is all you need to do.

Possible problems:
- pagemanager has its classes defined in the config file; in order to make extending easier, those config options will go away;

Did I miss anything? Do you guys see any problems I haven't in implementing this? Thanks, cheers!

lloy0076:
So long as the publish honours stuff in the configuration such as the 'base' (usually /admin but it is configurable).

[...]

PR #156, opened 2017-06-09
Enabled custom routing

eduardoarandah:
Hi! @tabacitu 

Im enabling custom routing for all packages. 

I created a config var "skip_all_backpack_routes" that disables all routes from being registered across all packages. 

It's disabled by default, so you only use it when the project grows. 

Custom routing will be in this file:

https://github.com/eduardoarandah/Base/blob/master/src/routes/backpack.php

Added this to readme:

## Custom Routing

By default, every component registers its routes automatically. 
If you want to disable this, in config/backpack/base.php
set to true: 

```'skip_all_backpack_routes' => true,```

Then add to app/Providers/RouteServiceProvider.php

```
public function map()
{
    ...
    
    $this->mapBackpackRoutes(); // <<--THIS LINE
    
}

protected function mapBackpackRoutes() <<--THIS FUNCTION
{
   Route::middleware('web')
        ->prefix(config('backpack.base.route_prefix', 'admin'))
        ->namespace('Backpack\PermissionManager\app\Http\Controllers')
        ->group(base_path('routes/backpack.php'));
}
```

## Permissions

When you have custom routing, you can easily restrict access with **PermissionManager** module

Example: restrict access to manage users

- Create a "manage-users" permission
- Add roles or users to that permission
- Add can:manage-users to the routes/backpack.php file like this

```CRUD::resource('user', 'UserCrudController')->middleware('can:manage-users');```

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,48 @@ $ php artisan migrate #generates users table (using Laravel's default migrations
 
 Please see [CHANGELOG](CHANGELOG.md) for more information what has changed recently.
 
+## Custom Routing
+
+By default, every component registers its routes automatically. 
+If you want to disable this, in config/backpack/base.php
+set to true: 
+
+```'skip_all_backpack_routes' => true,```
+
+Then add to app/Providers/RouteServiceProvider.php
+
+```
+public function map()
+{
+    ...
+    
+    $this->mapBackpackRoutes(); // <<--THIS LINE
+    
+}
+
+protected function mapBackpackRoutes() <<--THIS FUNCTION
+{
+   Route::middleware('web')
+        ->prefix(config('backpack.base.route_prefix', 'admin'))
+        ->namespace('Backpack\PermissionManager\app\Http\Controllers')
+        ->group(base_path('routes/backpack.php'));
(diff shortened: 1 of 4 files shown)


lloy0076:
https://github.com/Laravel-Backpack/Base/issues/131 will also address this...

tabacitu:
Hi @eduardoarandah ,

Sorry for the long ass delay. I've been "in the cave" preparing something super awesome that ties in nicely with Backpack. You'll love it, especially if you're a Laravel freelancer :-)

This as awesome attempt at solving the problem, thanks A LOT, it's very useful. I do have some ideas on how to improve on it, though. The bad thing about Backpack becoming popular is that we can only do this once. When we launch it, it will be very inconvenient for users to change how they do stuff.

I'll comment inline so it's easier to read. Very curious to find out what you and others think about the improvements.

Cheers!

tabacitu:
@eduardoarandah - just merged the other PR and tagging it soon, so we'll be able to do this by just creating the routes file ;-)

Cheers and thanks a lot for the PR!

PR #160, opened 2017-06-27
Backend login with permission

dfk7677:
This PR just adds a simple permission check to logging in the backend. It is meant for projects that want to keep the user function and the admin function separate, without having 2 different tables for each of these 2 groups.

It is completely optional to use and all changes are used if the 'permission_protection' option is set to 'true' in the backpack configuration.

P.S: I am not an experienced programmer.

### DIFF
--- a/src/BaseServiceProvider.php
+++ b/src/BaseServiceProvider.php
@@ -32,7 +32,8 @@ class BaseServiceProvider extends ServiceProvider
 
         // use the vendor configuration file as fallback
         $this->mergeConfigFrom(
-            __DIR__.'/config/backpack/base.php', 'backpack.base'
+            __DIR__.'/config/backpack/base.php',
+            'backpack.base'
         );
 
         $this->registerAdminMiddleware($this->app->router);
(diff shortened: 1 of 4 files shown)


tabacitu:
Hi @dfk7677 ,

Thanks a lot for the PR. I'm afraid I'm not going to merge this as-is, because the ```base.php``` config file has already gotten big enough. 

In my opinion, a customization like this would be better done with a middleware, say a ```role:admin``` or ```permission:admin``` like the [spatie/permission package recommends](https://github.com/spatie/laravel-permission#using-a-middleware). You can [reuse my code here](https://gist.github.com/tabacitu/acd7929cc9f60d104da1df4c39f0dad2) if you want.

I do think we should find a way to make this easier, but I think it's important to keep Backpack\Base completely decoupled from Backpack\PermissionManager. Most admin panels don't _need_ permission, or roles. So it's a layer of complexity they don't need. 

Please don't let this discourage you from submitting other PRs. Best practice is to open and issue first and see what a good course of action is, so you don't spend time on a PR that doesn't get merged.

Thanks again, cheers!

PR #166, opened 2017-07-06
Filter's class 'active' treatment

parabellumKoval:
Filter does not set the class as active when new value was selected.
When url has no hashes the upper line `var activeTab = $('[href="' + location.hash.replace("#", "#tab_") + '"]');` choose excess links and toggle off all active classes.
Changing to` location.hash && activeTab && activeTab.tab('show');`  solved the issue.


tabacitu:
Great, merged, thank you!

Issue #168, opened 2017-07-14
Security issue / data leakage of usernames


kruisdraad:
File views/vendor/backpack/base/inc/sidebar.blade.php
snippet:
`<img src="https://placehold.it/160x160/00a65a/ffffff/&text={{ mb_substr(Auth::user()->name, 0, 1) }}" class="img-circle" alt="User Image">`

All good and fine, but this leaks all the names of people to Placehold. in my opinion both a security as privacy issue.

victorlap:
You mean it leaks there initial? Because it is only one letter that is send to Placehold..

tabacitu:
Hi @kruisdraad ,

As @victorlap mentions, it's only one letter, so I don't think we need to worry about this. Not a leakage. In any case, this is "fixed" in [the upcoming version](https://github.com/Laravel-Backpack/Base/pull/137), where we pick up the gravatar if it exists. Don't know if we'll keep the gravatar feature, though :-)

Cheers!

PR #169, opened 2017-07-14
remove unwanted data leak, relates to issue #168

kruisdraad:
see issue for details, this PR is a proposal for a fix. The DIV style could be better, however i am not that good with HTML/DIV codes ;)

#168 (ed: this is a link to the issue)

### DIFF
--- a/src/resources/views/inc/sidebar.blade.php
+++ b/src/resources/views/inc/sidebar.blade.php
@@ -6,11 +6,10 @@
         <!-- Sidebar user panel -->
         <div class="user-panel">
           <div class="pull-left image">
-            <img src="https://placehold.it/160x160/00a65a/ffffff/&text={{ mb_substr(Auth::user()->name, 0, 1) }}" class="img-circle" alt="User Image">
+            <br>
           </div>
           <div class="pull-left info">
             <p>{{ Auth::user()->name }}</p>
-            <a href="#"><i class="fa fa-circle text-success"></i> Online</a>
           </div>
         </div>
         <!-- sidebar menu: : style can be found in sidebar.less -->


src/resources/views/inc/sidebar.blade.php
```diff
@@ -6,11 +6,10 @@
         <!-- Sidebar user panel -->
         <div class="user-panel">
           <div class="pull-left image">
-            <img src="https://placehold.it/160x160/00a65a/ffffff/&text={{ mb_substr(Auth::user()->name, 0, 1) }}" class="img-circle" alt="User Image">
+            <br>
```
lloy0076: I agree, in principle, with removing the call to placehold.it but I'm not sure replacing it with a break is the best solution.

tabacitu:
Hi @kruisdraad ,

Thanks a lot for the PR. As I mentioned in the issue, I don't think this is a security issue, or a leak. We only send one letter to Placehold.it - which I don't think is enough for _anything_ :-)

So in my opinion - not an issue. Better to have the placehold image. 
@lloy0076 , what do you think?

Cheers!

lloy0076:
A config option with some transparent gif would address any leakage to "placehold.it".

On the other hand, a single letter isn't too grand a leak (although it is technically an information leak).

That said, I'm on a stupidly slow connection at the moment and it *is* just one more external call that needs to be made (where each call is hit by a minimum of 1000ms latency)...

@tabacitu - I'm happy either way!

kruisdraad:
I also shows when people logged in and the refferer. It was very late when i saw the code, so i missed the 1 letter part, still for me i don't like it when every logging gets a call. Same reason i am localizing all the JS files on a static host ;)

tabacitu:
Hmm... I think that's overkill, myself. I might be wrong - it's been known to happen :-) - but I think we're better off having the placehold image. Feel free to remove it from your own projects, though - totally understand.

@lloy0076 - magic happens when you move to HTTP/2 :-) You no longer have to worry about having too many JS/CSS files loading - it's actually better to have many files rather than one big file, because they load asynchronously. I always use the [free CloudFlare plan](https://www.cloudflare.com/) as my DNS - you get HTTP/2, excellent caching, better security and some other nice features too.

tabacitu:
If anybody else stumbles upon this and feels I've been in the wrong, do tell :-)

PR #177, opened 2017-08-30
Remove route names

eduardoarandah:
when we want front-end users, adding Auth::routes(); to routes/web is impossible, because of route names overwritten

Look:
http://imgur.com/a/z5e9J

This change removes the conflict

### DIFF
--- a/src/routes/backpack/base.php
+++ b/src/routes/backpack/base.php
@@ -19,7 +19,28 @@ Route::group(
 function () {
     // if not otherwise configured, setup the auth routes
     if (config('backpack.base.setup_auth_routes')) {
-        Route::auth();
+        
+        //same routes as when calling Route::auth() but without route names
+        //why? 
+        //If you want front-end users, you can use Auth::routes(); without conflict in route names 
+        //check: vendor/laravel/framework/src/Illuminate/Routing/Router.php
+        //
+        // Authentication Routes...
+        Route::get('login', 'Auth\LoginController@showLoginForm');
+        Route::post('login', 'Auth\LoginController@login');
+        Route::post('logout', 'Auth\LoginController@logout');
+
+        // Registration Routes...
+        Route::get('register', 'Auth\RegisterController@showRegistrationForm');
+        Route::post('register', 'Auth\RegisterController@register');
+
+        // Password Reset Routes...
+        Route::get('password/reset', 'Auth\ForgotPasswordController@showLinkRequestForm');
+        Route::post('password/email', 'Auth\ForgotPasswordController@sendResetLinkEmail');
+        Route::get('password/reset/{token}', 'Auth\ResetPasswordController@showResetForm');
+        Route::post('password/reset', 'Auth\ResetPasswordController@reset');
+        //end Route::auth();
+        
         Route::get('logout', 'Auth\LoginController@logout');
     }
 


src/routes/backpack/base.php
```diff
@@ -19,7 +19,28 @@
 function () {
     // if not otherwise configured, setup the auth routes
     if (config('backpack.base.setup_auth_routes')) {
-        Route::auth();
+        
+        //same routes as when calling Route::auth() but without route names
```
lloy0076: Err, what about those who rely on the routes being name...and pointing at the below routes?

eduardoarandah: Yes, you're right. 

I guess it may help to just have in a FAQ something like: 

**If you plan on having both laravel and backpack auth routes**

copy this routes to routes/backpack/base.php and remove duplicate names like 'login'
<routes here>

tabacitu:
Ugh... you're right, this is a problem, and @lloy0076  you're right, this is a breaking change, but not that big - auth route  names are used _nowhere_ inside Backpack, so it's only if the developer has used them himself.

I tried to find an alternative solution so that we don't manually mention every auth route, but [the one solution I found](https://github.com/laravel/framework/issues/15058#issuecomment-259002543) also brings in a lot of route names like "backpack.auth." and I didn't like that.

I do think that the auth routes, even if they are just for the backend, should have names, so I've modified your PR a bit to add custom route names. We haven't used route names properly inside Backpack until now, but I think it's something we should do in the future. Using route names should allow developers to customize the URLs without affecting functionality.

Take a look at the routing convention:
- ```backpack.auth.login```
- ```backpack.auth.logout```
- ```backpack.account.password```
- ```backpack.account.info```
- etc

I think this makes for a pretty nice route naming convention - ```backpack``` prefix, then section name, then whatever that does:
screen_shot_2017-09-12_at_10_01_14

eduardoarandah:
There's a way of namespacing route names easily.

Just add `'as'=>'backpack.'` in the route group (with a dot in the end)

```
Route::group(
[
    'namespace'  => 'Backpack\Base\app\Http\Controllers',
    'middleware' => 'web',
    'prefix'     => config('backpack.base.route_prefix'),
    'as'=>config('backpack.base.route_prefix').'.'
],
function () {
...
});
```


Route names are VERY convenient for some reasons:

## Bash Grep

Let us have this alias

```
alias pa='php artisan'
alias par='php artisan route:list'
alias parg='php artisan route:list | grep '
```

when you have 100s of routes, it becomes easier to find them

`parg admin.auth`

## Permissions made easier

just match the permission name with route name and add `can:permissionname`
integrates nicely with backpack-permissions
and they are listed in order

```
Route::group(
[
    'namespace'  => 'admin',
    'middleware' => 'web',
    'prefix'     => 'admin',
    'as'     => 'admin.',    	
],
function () {
	Route::get('something', 'AbcController@something')->middleware('can:admin.abc.something');
});
```

Look at this **beauty**. 
All in order, all matching route names, OCD happy

imagen


## Route names break easily

**This is good!** 
because finding broken links are a pain

**lets see this code for the sidebar:**

```
@can('admin.sales.index')
<li><a href="{{ route('admin.sales.index') }}"><i class="fa fa-info-circle"></i> <span>Sales</span></a></li>
@endcan

@can('admin.setting')
<li><a href="{{ url('admin/setting') }}"><i class="fa fa-cog"></i> <span>Settings</span></a></li>
@endcan

```

If some route is missing, the first link will throw error, and we will see it while developing

the second one will remain as an obscure bug if it happens to be admin/setting**s** instead of admin/setting

imagen

OwenMelbz:
If you're making breaking changes anyway, just do it! Worth the grand total of maybe 5/10 mins to upgrade

PR #184, opened 2017-09-19
Fix Gravatar use with a custom authentication guard

RedAB:
When using Gravatar with a custom authentication guard, the application crashes as the user object is not obtained properly.


tabacitu:
Has since been fixed. Thanks for the PR @RedAB ! Cheers!

PR #185, opened 2017-09-19
Define 'admin' as a Middleware Group

jguyomard:
Hello,

I think using a middleware group (instead of an alias) is a better practice here. Why? Because it allows developers to add other middlewares in the group (programmatically or using the `app/Http/Kernel.php` file.

According to me, there is no breaking changes. 
I did some tests under Laravel 5.4 and Laravel 5.5.

### DIFF
--- a/src/BaseServiceProvider.php
+++ b/src/BaseServiceProvider.php
@@ -101,9 +101,9 @@ class BaseServiceProvider extends ServiceProvider
 
     public function registerAdminMiddleware(Router $router)
     {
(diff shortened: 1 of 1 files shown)


tabacitu:
Thanks a lot, @jguyomard ! I think you're totally right, it makes way more sense for it to be a middleware group, instead of a simple middleware.

I merged your changes into PR https://github.com/Laravel-Backpack/Base/pull/258, so I'm going to close this PR. Thanks again!

PR #188, opened 2017-09-24
uncdn: font-awesome and ionicon

alashow:
Added/Downloaded `font-awesome` and `ionicon` to public folder.


alashow:
https://github.com/Laravel-Backpack/CRUD/pull/1007

tabacitu:
@alashow I'm afraid we can't un-cdn Backpack at this time. Not until we figure out & implement a better solution for javascript dependency management. So that when we use newer versions of the JS/CSS files, they also get published & used.

This change would break a lot of people's apps, who have not republished the files. So we can't use this, unfortunately.

I took this opportunity to check if these are the latest version of those files and yes they are. So no need to update the CDN addresses.

Cheers!

PR #189, opened 2017-09-24
jquery-2.2.0 was removed from vendor/adminlete, changed to jquery-2.2.3

alashow:


lloy0076:
It's at `2.2.4` now 🙂

src/resources/views/layout.blade.php
```diff
@@ -102,7 +102,7 @@
 
     <!-- jQuery 2.2.0 -->
     <script src="https://code.jquery.com/jquery-2.2.0.min.js"></script>
-    <script>window.jQuery || document.write('<script src="{{ asset('vendor/adminlte') }}/plugins/jQuery/jQuery-2.2.0.min.js"><\/script>')</script>
+    <script>window.jQuery || document.write('<script src="{{ asset('vendor/adminlte') }}/plugins/jQuery/jQuery-2.2.3.min.js"><\/script>')</script>
```
lloy0076: Ideally if we're going to do this, we'd actually look through the plugins directory for something that "looked like" jQuery although this does work, I suppose.

alashow: or better: don't rely on AdminLTE and include it to backpack itself.

lloy0076: https://thecodingmachine.github.io/discovery/

I don't really recomment it - it just seems to add even more complexity.

alashow: I don't think there is a need for something like `discovery`. Just making sure the file is still there when updating AdminLTE or including it to backpack itself will do.

tabacitu:
Merged, thank you!

PR #191, opened 2017-09-27
Command-line installer

tabacitu:
FINALLY! Right? :-)

Instead of asking the developer to run a bunch of commands, ```php artisan backpack:base:install``` will run them all. 

This drastically reduces the number of steps needed to install Backpack\Base. Take a look at the output here: https://cl.ly/1b1v413s1q1T If we do this AND decide to use our own AdminUser class for authentication (that extends User), this will make installing Backpack\Base a two-step process:
```php
composer require backpack/base
php artisan backpack:base:install
```

That's the end goal here - for every Backpack package to be installed as easy as that, with that naming convention (```backpack:crud:install```, ```backpack:pagemanager:install```, etc). And every compatible Backpack package to follow this same standard.

@OwenMelbz , @lloy0076 what do you guys think so far? Command naming ok?

Thanks, cheers!

### DIFF
--- a/src/BaseServiceProvider.php
+++ b/src/BaseServiceProvider.php
@@ -9,7 +9,7 @@ use Route;
 class BaseServiceProvider extends ServiceProvider
 {
     protected $commands = [
-        'Backpack\Base\app\Console\Commands\Install',
+        Backpack\Base\app\Console\Commands\Install::class,
     ];
 
     /**
--- a/src/BaseServiceProvider.php
+++ b/src/BaseServiceProvider.php
@@ -8,6 +8,10 @@ use Route;
 
 class BaseServiceProvider extends ServiceProvider
 {
+    protected $commands = [
+        'Backpack\Base\app\Console\Commands\Install',
+    ];
+
     /**
      * Indicates if loading of the provider is deferred.
      *
@@ -97,6 +101,9 @@ class BaseServiceProvider extends ServiceProvider
                 $this->app->register('Backpack\Generators\GeneratorsServiceProvider');
             }
         }
+
+        // register the artisan commands
+        $this->commands($this->commands);
     }
 
     public function registerAdminMiddleware(Router $router)
--- a/src/app/Console/Commands/Install.php
+++ b/src/app/Console/Commands/Install.php
@@ -53,12 +53,14 @@ class Install extends Command
     /**
      * Run a SSH command.
      *
-     * @param  string  $command      The SSH command that needs to be run
-     * @param  boolean $beforeNotice Information for the user before the command is run
-     * @param  boolean $afterNotice  Information for the user after the command is run
-     * @return mixed                 Command-line output
+     * @param string $command      The SSH command that needs to be run
+     * @param bool   $beforeNotice Information for the user before the command is run
+     * @param bool   $afterNotice  Information for the user after the command is run
+     *
+     * @return mixed Command-line output
      */
-    public function executeProcess($command, $beforeNotice = false, $afterNotice = false) {
+    public function executeProcess($command, $beforeNotice = false, $afterNotice = false)
+    {
         if ($beforeNotice) {
             $this->info('### '.$beforeNotice);
         } else {
--- a/src/app/Console/Commands/Install.php
+++ b/src/app/Console/Commands/Install.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Backpack\Base\app\Console\Commands;
+
+use Illuminate\Console\Command;
+use Symfony\Component\Process\Process;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+
+class Install extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'backpack:base:install';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Require dev packages and publish files for Backpack\Base to work';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $this->info("Backpack\Base installation started. Please wait...");
+
+        $this->executeProcess("composer require backpack/generators --dev");
+        $this->executeProcess("composer require laracasts/generators:dev-master --dev");
+        $this->executeProcess("php artisan vendor:publish --provider='Backpack\Base\BaseServiceProvider'", "publishing configs, langs, views and AdminLTE files");
+        $this->executeProcess("php artisan vendor:publish --provider='Prologue\Alerts\AlertsServiceProvider'", "publishing config for notifications - prologue/alerts");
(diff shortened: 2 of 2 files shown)


bloggervista:
thanks :)

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $this->info("### Backpack\Base installation started. Please wait...");
+
+        $this->executeProcess('composer require backpack/generators --dev');
+        $this->executeProcess('composer require laracasts/generators:dev-master --dev');
```
lloy0076: Should the above two be made optional (but default to true)? Technically Backpack will work without them.

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
```
lloy0076: Should all the command output be translatable?

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->info("### Backpack\Base installation started. Please wait...");
+
+        $this->executeProcess('composer require backpack/generators --dev');
+        $this->executeProcess('composer require laracasts/generators:dev-master --dev');
+        $this->executeProcess("php artisan vendor:publish --provider='Backpack\Base\BaseServiceProvider'", 'publishing configs, langs, views and AdminLTE files');
+        $this->executeProcess("php artisan vendor:publish --provider='Prologue\Alerts\AlertsServiceProvider'", 'publishing config for notifications - prologue/alerts');
+        $this->executeProcess('php artisan migrate', "generating users table (using Laravel's default migrations)");
+
+        $this->info("### Backpack\Base installation finished.");
```
lloy0076: Translatable?

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Run a SSH command.
+     *
+     * @param string $command      The SSH command that needs to be run
+     * @param bool   $beforeNotice Information for the user before the command is run
```
lloy0076: These are at best `mixed`.

Honestly, I'd make them `string` and default them to `null`. And check for `if ($isset($beforeNotice) { /* display it */ }`.

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+        if ($beforeNotice) {
+            $this->info('### '.$beforeNotice);
+        } else {
+            $this->info('### Running: '.$command);
+        }
+
+        $process = new Process($command);
+        $process->run(function ($type, $buffer) {
+            if (Process::ERR === $type) {
+                echo '... > '.$buffer;
```
lloy0076: Why the `echo` and not an appropriate `info` or `line` or even `error`?

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+            } else {
+                echo 'OUT > '.$buffer;
```
lloy0076: ditto

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+            }
+        });
+
+        // executes after the command finishes
```
lloy0076: Really? 😈

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+        if (!$process->isSuccessful()) {
+            throw new ProcessFailedException($process);
+        }
+
+        if ($afterNotice) {
```
lloy0076: See comments re type of this argument.

src/BaseServiceProvider.php
```diff
@@ -8,6 +8,10 @@
 
 class BaseServiceProvider extends ServiceProvider
 {
+    protected $commands = [
+        'Backpack\Base\app\Console\Commands\Install',
```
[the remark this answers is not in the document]
bloggervista: I would also say ::class is better as it is IDE friendly

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+     * @var string
+     */
+    protected $signature = 'backpack:base:install';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Require dev packages and publish files for Backpack\Base to work';
```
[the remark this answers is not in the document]
alashow: Don't think so. 
I don't think Laravel has options for translating command descriptions neither.

src/BaseServiceProvider.php
```diff
@@ -97,6 +101,9 @@ public function register()
                 $this->app->register('Backpack\Generators\GeneratorsServiceProvider');
             }
         }
+
+        // register the artisan commands
```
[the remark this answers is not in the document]
tabacitu: True, but it's a Laravel-method, I think renaming it would just confuse. I say let's leave the comment, it provides enough clarity.

src/app/Console/Commands/Install.php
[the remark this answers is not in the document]
tabacitu: Agreed. I don't think Laravel's command are translatable either. Since the only people who will see this are developers, English is a must. We don't even have docs in different languages, so I say English should be the only supported language for commands.

src/app/Console/Commands/Install.php
tabacitu: Yup, you're right. They could be optional. I thought including them in require-dev will do no harm (they won't be present in production), but it should make it more intuitive for the user: "_of course the generators are included automagically on dev_". You think it's confusing rather than helpful?

src/app/Console/Commands/Install.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public function handle()
+    {
+        $this->info("### Backpack\Base installation started. Please wait...");
+
+        $this->executeProcess('composer require backpack/generators --dev');
+        $this->executeProcess('composer require laracasts/generators:dev-master --dev');
+        $this->executeProcess("php artisan vendor:publish --provider='Backpack\Base\BaseServiceProvider'", 'publishing configs, langs, views and AdminLTE files');
+        $this->executeProcess("php artisan vendor:publish --provider='Prologue\Alerts\AlertsServiceProvider'", 'publishing config for notifications - prologue/alerts');
+        $this->executeProcess('php artisan migrate', "generating users table (using Laravel's default migrations)");
```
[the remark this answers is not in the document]
tabacitu: They end up executing the same commands, in the end, I just found this wau more consistent - all command are run using ```executeProcess()``` which shows the same output.

src/app/Console/Commands/Install.php
tabacitu: Not needed, IMHO.

src/app/Console/Commands/Install.php
lloy0076: The more that is required, the more likely something will break - it's up to you.

PR #196, opened 2017-10-24
Backpack register variable

joaojoyce:
Hi,

Please consider this simple change. 
``````
    'registration_open' => env('BACKPACK_OPEN_REGISTRATION',false),
``````

Open / Closed registration should be configured by explicitly setting an environment variable. 

This will avoid the possibility of someone unintentionally opening registration for every one by modifying the APP_ENV variable for an unrelated reason.

### DIFF
--- a/src/config/backpack/base.php
+++ b/src/config/backpack/base.php
@@ -47,7 +47,7 @@ return [
     |
     */
 
-    'registration_open' => env('BACKPACK_REGISTRATION_OPEN', true),
+    'registration_open' => env('BACKPACK_REGISTRATION_OPEN', false),
 
     /*
     |--------------------------------------------------------------------------
--- a/src/config/backpack/base.php
+++ b/src/config/backpack/base.php
@@ -47,7 +47,7 @@ return [
     |
     */
 
-    'registration_open' => env('BACKPACK_OPEN_REGISTRATION',false),
+    'registration_open' => env('BACKPACK_OPEN_REGISTRATION', false),
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


joaojoyce:
I believe it would be safer for this default to be "false". There are a lot of sites that only need a single admin user.

In any case I'm totally fine with it being true to avoid breaking stuff.

lloy0076:
@joaojoyce - as someone concerned with "secure by default out of the box", I agree; however, AFAICT the default is for the registration to be open both for Backpack and even Laravel so it's a change that may surprise some people.

tabacitu:
Hi guys,

In order to keep this backwards-compatible, I agree, let's make it ```true``` by default.

Thanks for the PR, @joaojoyce . This ENV variable makes a lot of sense to me, and I think it will make it easier for a lot of people to test their apps on different installations (localhost, staging, production). Keep 'em coming :-)

Cheers!

tabacitu:
Whoow. Just realised this is a MAJOR security flaw if people do a ```composer update``` to their production apps. So no, we can't merge this with default ```true``` as it will expose registration in production for a lot of apps. We can only merge it by default with ```false```. Which makes this a breaking change. I agree to it, but this PR has to wait until the next point update, I'm afraid.

:-) Pfiew.

jguyomard:
Maybe `env('BACKPACK_REGISTRATION_OPEN', env('APP_ENV') === 'local')` if you don't want any bc.

tabacitu:
@jguyomard totally. Merged as you suggested - will be available shorty with a ```composer update```.

Cheers, thank you for your patience during the back-and-forth!

PR #197, opened 2017-11-02
add html locale before header

trakons:
I think it can be useful to have the same locale as the page translation


tabacitu:
Sure, sound like a good addition, thank you!

PR #198, opened 2017-11-06
Update BaseServiceProvider.php

rvpalivoda:


tabacitu:
Hi @rvpalivoda ,

Thanks a lot for the PR. The fix had been plugged using ```app\Console\Commands\Install::class```, but I like your solution better - full namespace makes the app clearer.

Thanks, merged, cheers!

Issue #200, opened 2017-11-06
Bugs for fresh install backpack base ^=^


MDEvo:
# Bug report

On fresh install Laravel 5.5 when run 

`composer require backpack/base`

will throw this error:
```
  [ReflectionException]
  Class Backpack\Base\Backpack\Base\app\Console\Commands\Install does not exist
```

There is repetition:

> Backpack\Base\Backpack\Base\app\Console\Commands\Install should be Backpack\Base\app\Console\Commands\Install

MDEvo:
Solved by changing:

> protected $commands = [
>         \Backpack\Base\app\Console\Commands\Install::class,
> ];

 on BaseServiceProvider.php

fhusquinet:
Same issue with Laravel 5.5, looks like the command has the namespace automatically prepending to it throwing the exception.

xcerx:
@fhusquinet you can work around by fix Backpack\Base\BaseServiceProvider.php

from ...
```php
protected $commands = [
        Backpack\Base\app\Console\Commands\Install::class,
    ];
```

to ...
```php
protected $commands = [
        app\Console\Commands\Install::class,
    ];
```

lloy0076:
@tabacitu There's various iterations of this floating about - you have any idea of what these are?

tabacitu:
Hi guys, this has been fixed in a matter of hours, but I didn't get to see this issue too. I've just tested the installation of the current version and it works fine, so I'll close the issue.

Please reopen the issue if you're still having this problem.

Cheers!

PR #202, opened 2017-11-07
Fix cannot run php artisan package:discover

xcerx:
There is error when run command "php artisan package:discover", because of Class Backpack\Base\Backpack\Base\app\Console\Commands\Install does not exist


xcerx:
fix #200

tabacitu:
Thanks for the PR, @xcerx . This has been fixed in matter of hours, probably with a similar PR, so I'll close this. Keep 'em coming though :-) Cheers!

PR #208, opened 2017-11-13
get $user object and use it, otherwise not working with customized guard

kiddtang:
Still using Auth Facade to get $user object while user passing in...

```
return Gravatar::fallback('https://placehold.it/160x160/00a65a/ffffff/&text='.Auth::user()->name[0])->get($user->email);
```


tabacitu:
Yup, you're totally right @kiddtang . Thanks for the PR, merged!

PR #215, opened 2017-11-16
Using guard() method to access the current guard in MyAccountController

### LINKED ISSUE #210: MyAccountController should use single $user object
I am using non-default guard for admin, default guard for user.

`MyAccountController` using user object by `Auth::user()` and called every single function.

Why don't just put in a constructor so i can extend and override the function?

Thanks.

pavoltanuska:
So we are able to override the Guard to a custom one in one place instead of all over the Controller.

It's (afaik) the standard way to do this - the default Login/Register Traits are doing it like this, so should we :)

Should also fix #210

### DIFF
--- a/src/app/Http/Controllers/Auth/MyAccountController.php
+++ b/src/app/Http/Controllers/Auth/MyAccountController.php
@@ -24,7 +24,7 @@ class MyAccountController extends Controller
     public function getAccountInfoForm()
     {
         $this->data['title'] = trans('backpack::base.my_account');
-        $this->data['user'] = Auth::user();
+        $this->data['user'] = $this->guard()->user();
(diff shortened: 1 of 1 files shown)


lloy0076:
Actually, why aren't we just using the actual traits and implementing them...is the bigger question.

tabacitu:
Totally, thank you @pavoltanuska !

@lloy0076 - what traits are you referring to? The traits I see in LoginController, RegisterController, ResetPasswordController, ForgotPasswordController all have a bunch of methods we wouldn't need in MyAccountController.

Cheers!

lloy0076:
I must have been half asleep when writing that comment 😄

Issue #222, opened 2017-11-23
Publishing Adminlte assets not working in not default base_path


yoramdelangen:
I have an installation containing 2 different base_paths/app_paths. Instead of the default `app/` I have `backend/app/` and `frontend/app/` (so the base path will be set to `backend/` and `frontend/`. Both base_paths are booted correctly. But because the base path is different and one level deeper then the usual it causes problems with publishing the adminlte assets (because it is using the `base_path()` function).

Easly to reproduce is setting the base_path to `backend/app` and running the command:
`php artisan vendor:publish --tag=adminlte`

Fixed it for now temporary by removing the `base_path()` function and calculate the path from current directory. In this way its more accurate thats in the vendor directory instead of the `base_path()` function.

Code for pull request to fix:
```
// calculate the path from current directory to get the vendor path
$vendorPath = __DIR__.'/../../..';

// publish public AdminLTE assets
$this->publishes([$vendorPath.'/almasaeed2010/adminlte' => public_path('vendor/adminlte')], 'adminlte');

// publish public Gravatar assets
$this->publishes([$vendorPath.'/creativeorange/gravatar/config' => config_path()], 'gravatar');
```

Latest Backpack, Laravel 5.5.21, PHP 7.1.8

tabacitu:
Closing this - I believe this fixes it https://github.com/Laravel-Backpack/Base/pull/223 - someone was very kind to submit a PR for it :-)

Thanks, cheers!

PR #223, opened 2017-11-23
Calculate vendor path from current BaseServiceProvider location #222

yoramdelangen:
See issue #222

### DIFF
--- a/src/BaseServiceProvider.php
+++ b/src/BaseServiceProvider.php
@@ -139,10 +139,13 @@ class BaseServiceProvider extends ServiceProvider
         // publish public Backpack assets
         $this->publishes([__DIR__.'/public' => public_path('vendor/backpack')], 'public');
 
+        // calculate the path from current directory to get the vendor path
+        $vendorPath = __DIR__.'/../../..';
+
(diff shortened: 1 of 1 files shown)


src/BaseServiceProvider.php
```diff
@@ -139,10 +139,13 @@ public function publishFiles()
         // publish public Backpack assets
         $this->publishes([__DIR__.'/public' => public_path('vendor/backpack')], 'public');
 
+        // calculate the path from current directory to get the vendor path
+        $vendorPath = __DIR__.'/../../..';
```
jguyomard: `$vendorPath = dirname(__DIR__, 3);`

(ok, it doesn't change anything, it's just more elegant :smile:)

yoramdelangen: Agreed! fixed!

tabacitu:
You're totally right, this IS a better way to go, thank you @yoramdelangen for the PR, and @jguyomard for confirming the solution works.

Also tested - works like a charm. Merged, will be available to everyone in the next patch update (today or later this week). 

Thanks, cheers!

PR #229, opened 2017-12-25
Update sidemenu.blade.php

AbbyJanke:
Add user's name and avatar preview in the account setting sidebar similar to how it is shown within AdminLTE Demo:

https://adminlte.io/themes/AdminLTE/pages/examples/profile.html

### DIFF
--- a/src/resources/views/auth/account/sidemenu.blade.php
+++ b/src/resources/views/auth/account/sidemenu.blade.php
@@ -1,4 +1,10 @@
 <div class="box">
+    <div class="box-body box-profile">
+
+	    <img class="profile-user-img img-responsive img-circle" src="{{ backpack_avatar_url(Auth::user()) }}">
+	    <h3 class="profile-username text-center">{{ Auth::user()->name }}</h3>
+
+	</div>
 	<ul class="nav nav-pills nav-stacked">
 
 	  <li role="presentation"


src/resources/views/auth/account/sidemenu.blade.php
```diff
@@ -1,4 +1,10 @@
 <div class="box">
+    <div class="box-body box-profile">
+
```
lloy0076: What's that extra line for?

src/resources/views/auth/account/sidemenu.blade.php
```diff
@@ -1,4 +1,10 @@
...
+	    <img class="profile-user-img img-responsive img-circle" src="{{ backpack_avatar_url(Auth::user()) }}">
```
lloy0076: And if that `Auth::user()` returns `null`?

src/resources/views/auth/account/sidemenu.blade.php
AbbyJanke: Following the overall style of the rest of the file.

src/resources/views/auth/account/sidemenu.blade.php
AbbyJanke: If you look at sidebar.blade.php (which is where i copied the line from) it is exactly the same so look at @tabacitu for that answer.

lloy0076: If it's `null` we'll get a PHP error; I guess seeing we're meant to be logged in here it won't ever be `null`.

tabacitu:
Yup, totally agree with this. Just added another <hr> for style. Merged, thank you guys, cheers!

PR #233, opened 2018-01-15
[Feature][1.0][WIP] Support AdminLTE original Auth pages style with some changes to blades structs

iMokhles:
separated head, main_header and scripts to new blades, added support for adminlte original login page style for auth blades

example of auth pages

Register Page

capture d ecran 2018-01-15 a 03 14 53

Login Page

capture d ecran 2018-01-15 a 03 14 41

Email Page
capture d ecran 2018-01-15 a 03 14 46

### DIFF
--- a/src/public/overlays/backpack.bold.css
+++ b/src/public/overlays/backpack.bold.css
@@ -48,3 +48,14 @@ body {
 .pace .pace-activity {
 	display: none;
 }
+
+.login-page, .register-page {
+	background: #ecf0f5;
+}
+
+.login-footer {
+	position: absolute;
+	text-align: center;
+	bottom: 5px;
+	width: 100%;
+}
--- a/src/resources/views/auth/login.blade.php
+++ b/src/resources/views/auth/login.blade.php
@@ -1,66 +1,90 @@
 @extends('backpack::layout')
 
-@section('content')
-    <div class="row">
-        <div class="col-md-8 col-md-offset-2">
-            <div class="box box-default">
-                <div class="box-header with-border">
-                    <div class="box-title">{{ trans('backpack::base.login') }}</div>
+@section('after_styles')
+    <link rel="stylesheet" href="{{ asset('vendor/adminlte/') }}/plugins/iCheck/square/blue.css">
+    <link rel="stylesheet" href="{{ asset('vendor/adminlte/') }}/dist/css/auth.css">
+@endsection
+
+@section('body_attributes')
+    login-page
+@endsection
+
+@section('login-box')
+
+    <div class="login-box">
+        <div class="login-logo">
+            <a href="{{ url('') }}">{!! config('backpack.base.logo_lg') !!}</a>
+        </div>
+        <!-- /.login-logo -->
+        <div class="login-box-body">
+            <p class="login-box-msg">{{ trans('backpack::base.login') }}</p>
+
+            <form role="form" method="POST" action="{{ route('backpack.auth.login') }}">
+                {!! csrf_field() !!}
+
+                <div class="form-group has-feedback {{ $errors->has('email') ? 'has-error' : '' }}">
+                    <input type="email" name="email" class="form-control" value="{{ old('email') }}"
+                           placeholder="{{ trans('backpack::base.email_address') }}">
+                    <span class="glyphicon glyphicon-envelope form-control-feedback"></span>
+                    @if ($errors->has('email'))
+                        <span class="help-block">
+                            <strong>{{ $errors->first('email') }}</strong>
+                        </span>
+                    @endif
                 </div>
-                <div class="box-body">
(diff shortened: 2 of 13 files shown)


tabacitu:
I like this! I think it's a lot better than the pages we have now, thank you @iMokhles !

This is a breaking change, which is good - it means we can take our time to do it right :-) 

1) I really don't have much to say in terms of design - I think it's pretty much spot on. I think you've done a very good job, @iMokhles ;
2) I like how you split ```head``` and ```scripts```; long overdue :-) maybe it would be more consistent to rename ```scripts``` to ```bottom```? head and bottom - idk what do others think?
3) I'd make the background lighter on these pages, I think the current one is pretty dark; In UI, dark colors usually trigger bad feelings, and we wouldn't want the first interaction with the admin panel to be a "dark" one :-)  Maybe the same color as the background across the admin panel? [Difference screenshot here](https://cl.ly/2F2N241X2K13). This would be a simple change in ```backpack.bold.css```:
```
.login-page, .register-page {
   background: #ecf0f5;
}
```
4) Looks like the ```layout.blade.php``` file, which is an important file that a lot of people overwrite, got difficult to understand in the process; mostly because of the ```login-box```; instead of the general layout file it used to be, it now has a bunch of conditionals to test if it's the login page. This would make sense for a file that's primarily used for the login page, but it's not. It's primarily used for _everything else_. Login is just one page, of the many it is used for. So I don't really like the conditionals there. I'm wondering - **wouldn't it be cleaner to have two layout files?** 
   - ```layout.blade.php``` which is the default across the admin; has menu and sidebar;
   - ```layout_logged_out.blade.php```, which we can use for login, register, forgot password (maybe even error pages)? 

Basically (4) is the only actual problem I see, the others are just nitpicking :-)

What do you think about these changes @iMokhles ? 
@Laravel-Backpack/veterans what do you think about the new screens?

Ghitu:
1. the design in very nice
2. If we only use it for JavaScript we should keep the "scripts" name
3. +1 for lighter background
4. I think splitting is a good idea (layout_logged_out.blade.php => layout_guest.blade.php?)
5. My suggestion is a discrete text and link in the footer

iMokhles:
thanks for your comments guys, actually it based on my own backpack setup yeah i was thinking a lot about splitting layouts and i have done it once in one of my projects with backpack also ( so it won't be hard to split it now and i agree with you i feel good with it but i guess i was lazy to split it =_=" )
here's an example how i split my layouts.
capture d ecran 2018-01-18 a 13 06 21

about dark ui of course it will be better with the current backpack background color, i left the original as it's from AdminLTE site so of course it's great to have it lighter ;)

i agree with @Ghitu about scripts because it contains only scripts for now but we can change it if you would add anything else later rather than Javascript ?

yeah about the footer i left it for suggests because i hadn't idea where it should be present at the bottom of the page or just under the login box ?

eduardoarandah:
I always put sidebar items in a separate file called “menu”

Much easier to search “menu” in sublime than “vendor inc sidebar” 

Sidebar items is also the only file to change when adding or removing functionality.

iMokhles:
oh i forgot i had done something like MenuManager in my IMBaaS project
[IMBaaS/sidebar.blade.php#L24](https://github.com/iMokhles/IMBaaS/blob/master/resources/views/vendor/backpack/base/inc/sidebar.blade.php#L24)
[sidemenu_classes/migration](https://github.com/iMokhles/IMBaaS/blob/03617100084b24a0fdd01498fd9bf4ebfa2c8ef2/database/migrations/2017_07_19_154500_create_imbaas_database.php#L203)

so i'm building menu through Admin page through edit forms like anything else.

tabacitu:
Looks like we have consensus! Great! :-)

Modifs roundup:
[ ] separate ```layout_guest.blade.php``` (definitely a better name than layout_logged_out)
[ ] keep ```scripts``` name
[ ] add discreet footer (i'd say at the bottom of the page, light grey or smth; the link is what's important)

Separately (different issues I think, but we should launch them at the same time):
[ ] menu.blade.php

iMokhles:
Hey thanks @tabacitu for your updates
i have applied the suggests and added some changes from my side

[] separated footers
[] modified menu.blade ( to remove register and login buttons ) we just need logout button there after our changes ( normally ) users won't be able to see the menu if they are not logged in

tabacitu:
Hi @iMokhles ,

Thank you so much for the work you’ve put in here. Excellent.

I did not merge this PR per-se - so it’ll look “closed”, but I did take your files one-by-one and applied all changes you’ve done. So it’s merged, even though it doesn’t look like it :-) I just had to do it this way because there have been too many changes since your PR, and there were too many merge conflicts.

It’s now separated as PR https://github.com/Laravel-Backpack/Base/pull/330 - take a look and tell me what you think :-)

Thanks again for this @iMokhles and @lloy0076 ! We’re finally getting prettier auth pages and a ```layout_guest``` without sidebar. Whoop-whoop!

PR #237, opened 2018-01-25
Avoid calling Auth multiple times.

AbbyJanke:
Instead of recalling the `Auth::user()` instance use the `$user` variable which already calls the Auth method within the controller.

This is a fix/minor change for my original PR #229


tabacitu:
Thank you @AbbyJanke ! Valid point.

I'm going to close this in favour of https://github.com/Laravel-Backpack/Base/pull/165 - which brings the option of using a custom guard across Backpack. In order to pull in the correct info, the guard needs to always be used. So that PR replaces your ```$user``` calls with ```backpack_auth()```, which uses the correct guard.

Cheers!

PR #246, opened 2018-02-13
Auth\Login: allow username customization

kichetof:
If you change your User model to use another field that email to authenticate, you can't use BackPack.
(In my case, I use Adldap2 to login with AD username)
You need to update `login.blade.php` to allow `username` instead of `email` field.

### DIFF
--- a/src/app/Http/Controllers/Auth/LoginController.php
+++ b/src/app/Http/Controllers/Auth/LoginController.php
@@ -9,6 +9,7 @@ use Illuminate\Http\Request;
 class LoginController extends Controller
 {
     protected $data = []; // the information we send to the view
+    protected $user_auth_key = 'email';
 
     /*
     |--------------------------------------------------------------------------
@@ -33,6 +34,8 @@ class LoginController extends Controller
     {
         $this->middleware('guest', ['except' => 'logout']);
 
+        $this->user_auth_key = config('backpack.base.user_auth_key', 'email');
+
         // ----------------------------------
         // Use the admin prefix in all routes
 
@@ -48,6 +51,16 @@ class LoginController extends Controller
         $this->redirectAfterLogout = property_exists($this, 'redirectAfterLogout') ? $this->redirectAfterLogout
             : config('backpack.base.route_prefix', 'admin');
         // ----------------------------------
+    }
+
+    /**
+     * Return custom username for authentication.
+     *
+     * @return string
+     */
+    public function username()
+    {
+        return $this->user_auth_key;
     }
 
     // -------------------------------------------------------
--- a/src/config/backpack/base.php
+++ b/src/config/backpack/base.php
@@ -81,6 +81,9 @@ return [
     // Fully qualified namespace of the User model
     'user_model_fqn' => '\App\User',
 
+    // Username key for authentication
+    'user_auth_key' => 'email',
+
     // What kind of avatar will you like to show to the user?
     // Default: gravatar (automatically use the gravatar for his email)
     // Other options:


tabacitu:
@kichetof ,

Thanks for the PR. Seems like a good improvement to me. But wouldn't this also need some changes in the blade files and maybe validation? If the developer only changes the config option, will it work?

Thanks again!

iMokhles:
yes @tabacitu he mentioned that 
`You need to update login.blade.php to allow username instead of email field.`
so we can set configuration variables for it with default value ( `email` ) and devs change it to ( `username` ) from the `backpack.base` config file

kichetof:
In my case, I only edit `login.blade.php` to change `email` to `username` _(and input type='email' to type='text')_ (nothing in validation).

I intentionally didn't update the blades files in the PR. If you change the way to connect with different username, you need to know how to change register / password reset methods or disable it (as I do; registered user in AD server without option to edit password in web app).

If you want to allow dev to only change `backpack.base.user_auth_key` and everything works, we need to update more files and add more customization (translation, input type, ..)

tabacitu:
Hmm... I don't know... I think we should either 

(A) include a complete solution (change ```email``` to ```username``` in one spot, and it just works)

or 

(B) provide no help for this at all 

While a part-solution might help some users, I don't think it's a good choice in the long run, since it would certainly annoy others that don't understand what they need to do. It would bring a lot of questions and support requests. Again - I really appreciate the PR, thank you for it, I just want to see what's the best way we can go further. Before we do, let's ask some more people what they think.

Thank you!

tabacitu:
@Laravel-Backpack/veterans do you think Backpack/Base should come with the ability to change the login from ```email & password``` to ```username & password```? Have you ever needed this?

lloy0076:
Honestly I've found that I end up changing/reimplementing or meddling with Laravel's default authentication as either the first or second step in any project.

niladam:
This could be a game changer. Being able to pick and change what you use as username could improve the adoption rate (which is really high, I know); 

A good idea would be to actually provide the user with a multitude of options (ie: phone number, one time code and so on) would be a really great addition. However, I'm not sure if this should be in backpack and not Laravel's auth..

tabacitu:
Let's see if we can make this thing happen! :-) I merged this PR into https://github.com/Laravel-Backpack/Base/pull/257 and expanded it. Let's move the conversation there please.

**Warning:** I'm not 100% sure this is a feature we want/need yet. Not if we can't provide a reasonable solution. And there are a few problems I expect we'll encounter, but I do think it's worth a shot. That's why I merged it into that other PR and worked on it.

@lloy0076 what did you end up changing in the authentication, every time? I'm curious - might be a pain point the Backpack can solve?

@niladam I can't say I agree that this feature would increase adoption :-) Backpack has been around for ~2 years and this is the first time I hear about people wanting to change ```email``` to ```username``` for login. So I don't think a lot of people need it. Plus, using ```username``` for login is an antiquated process IMHO. It's bad UX. As an admin/user, it's more difficult to remember your username, then your email. Plus, your email is _already_ a unique field, while your preferred ```username``` could be grabbed by someone else before you do. I think ```username``` only makes sense in anonymous communities. Not in admin panels. But I do understand that porting old systems to Backpack might need to use the old ```username```. That's why I agree to explore this option. 

I think other fields (phone number, one-time-code etc) are definitely beyond the scope of Backpack. They can be so varied, that we have no way of helping the developer there. We can only make it easy for him to change the code. And I think that's where Backpack shines - it's dead-simple to change stuff.

PR #251, opened 2018-02-28
[Feature][0.9][Merged] Alert messages can now have HTML content

tabacitu:
So that you can do things like:

```
<strong>Something happened</strong><br>
Details about that something, that nobody will probably ever read.
```

This is NOT a breaking change, it's backwards-compatible, but we should make sure we don't use this feature ourselves in the code, until it's adopted. For this reason, I think it should be launched with the 3.4 update.

### DIFF
--- a/src/public/backpack.base.css
+++ b/src/public/backpack.base.css
@@ -841,6 +841,13 @@ body[class^='skin-purple'] .pagination>.active>span:hover {
     border-color: #f8d053;
   }
 
+  /*    FLEXBOX HELPERS */
+
+  .display-flex-wrap {
+    display: flex;
+    flex-wrap: wrap;
+  }
+
   /*    BORDER RADIUS HELPERS */
 
   .b-rad-sm {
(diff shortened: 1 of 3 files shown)


src/public/backpack.base.css
```diff
@@ -841,6 +841,13 @@ body[class^='skin-purple'] .pagination>.active>span:hover {
     border-color: #f8d053;
   }
 
+  /*    FLEXBOX HELPERS */
```
tabacitu: This has NOTHING to do with the current PR, but it's needed, so I say let's leave it.

lloy0076: Needed by what though?

src/resources/views/inc/alerts.blade.php
```diff
@@ -13,7 +13,7 @@
             $(function(){
               new PNotify({
                 // title: 'Regular Notice',
-                text: "{{ $message }}",
+                text: "{!! str_replace('"', "'", $message) !!}",
```
lloy0076: You sure you've checked this?

What if I have:

```
$message = '"';
```

tabacitu: Yup. Just double-checked to be sure: https://cl.ly/3C2V1s0i1Y26

tabacitu:

PR #257, opened 2018-03-06
[Feature][0.9][Merged] Allow login with username

tabacitu:
Merged PR https://github.com/Laravel-Backpack/Base/pull/246 and made a few minor changes (variable naming, necessary view changes). 

**Warning: I'm not 100% this is a feature we need/want yet. But I think it's worth a shot. That's why I finished it.**

### DIFF
--- a/src/app/Http/Controllers/Auth/ResetPasswordController.php
+++ b/src/app/Http/Controllers/Auth/ResetPasswordController.php
@@ -32,6 +32,10 @@ class ResetPasswordController extends Controller
     {
         $this->middleware('guest');
 
+        if (!backpack_users_have_email()) {
+            abort(403, trans('backpack::base.no_email_column'));
+        }
+
         // where to redirect after password was reset
         $this->redirectTo = property_exists($this, 'redirectTo') ? $this->redirectTo : config('backpack.base.route_prefix', 'admin').'/dashboard';
     }
--- a/src/app/Http/Controllers/Auth/ResetPasswordController.php
+++ b/src/app/Http/Controllers/Auth/ResetPasswordController.php
@@ -33,7 +33,7 @@ class ResetPasswordController extends Controller
         $this->middleware('guest');
 
         if (!backpack_users_have_email()) {
-            abort(403, trans('backpack::base.no_email_column'));
+            abort(501, trans('backpack::base.no_email_column'));
         }
 
         // where to redirect after password was reset
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -17,6 +17,33 @@ if (!function_exists('backpack_url')) {
     }
 }
 
+if (!function_exists('backpack_authentication_column')) {
+    /**
+     * Return the username column name.
+     * The Laravel default (and Backpack default) is 'email'.
+     *
+     * @return string
+     */
+    function backpack_authentication_column()
+    {
+        return config('backpack.base.authentication_column', 'email');
+    }
+}
+
+if (!function_exists('backpack_users_have_email')) {
+    /**
+     * Check if the email column is present on the user table.
+     *
+     * @return string
+     */
+    function backpack_users_have_email()
+    {
+        $user_model_fqn = config('backpack.base.user_model_fqn');
+        $user = new $user_model_fqn();
+        return \Schema::hasColumn($user->getTable(), 'email');
+    }
+}
+
 if (!function_exists('backpack_avatar')) {
     /**
      * Returns the avatar URL of a user.
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -55,19 +55,19 @@ if (!function_exists('backpack_avatar')) {
      */
     function backpack_avatar_url($user)
     {
-        $placehold = 'https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0];
+        $placeholder = 'https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0];
 
         switch (config('backpack.base.avatar_type')) {
             case 'gravatar':
                 if (backpack_users_have_email()) {
                     return Gravatar::fallback('https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0])->get($user->email);
                 } else {
-                    return $placehold;
+                    return $placeholder;
                 }
                 break;
 
             case 'placehold':
-                return $placehold;
+                return $placeholder;
                 break;
 
             default:
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -27,13 +27,19 @@ if (!function_exists('backpack_avatar')) {
      */
     function backpack_avatar_url($user)
     {
+        $placehold = 'https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0];
+
         switch (config('backpack.base.avatar_type')) {
             case 'gravatar':
-                return Gravatar::fallback('https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0])->get($user->email);
+                if (config('backpack.base.authentication_column')=='email') {
+                    return Gravatar::fallback('https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0])->get($user->email);
+                } else {
+                    return $placehold;
+                }
                 break;
 
             case 'placehold':
-                return 'https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0];
+                return $placehold;
                 break;
 
             default:
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -58,7 +58,7 @@ if (!function_exists('backpack_avatar')) {
 
         switch (config('backpack.base.avatar_type')) {
             case 'gravatar':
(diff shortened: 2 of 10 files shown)


tabacitu:
Let's sum it up.

### What does this PR do?

It allows the developer to quickly switch from using ```email``` for login to using ```username``` or any other column, just by changing some variables in ```config/backpack/base.php```:

```diff
    // Username column for authentication
    // The Backpack default is the same as the Laravel default (email)
    // If you need to switch to username, you also need to create that column in your db
-    'authentication_column' => 'email',
+    'authentication_column' => 'username',
-    'authentication_column_name' => trans('backpack::base.email_address'),
+    'authentication_column_name' => 'Username',
```

Result:

screen shot 2018-03-06 at 09 32 39

The login will "_just work_". However:

### Problems:

(1) [the "_password reset_" screen](https://cl.ly/370l2W0g163C) does NOT ask for ```username```; it will still ask for ```email```;
(2) [the "_my account_" screen](https://cl.ly/223q3y0m460f) does NOT allow the admin to change his ```username```;  only ```name``` and ```email```;
(3) [the "_register_" screen](https://cl.ly/3j012t2z332T) does NOT show that custom column; only ```name```, ```email```, ```password``` and ```confirm_password```;

### My thoughts on the problems above

Problem (1). On the one hand, of course you plug in your email, that's where it will send the reset link, after all. On the other hand, it highlights that _```email``` would still be a required column in the DB_. Which is not the case for most systems that need ```username```, I think. If you have ```email``` and it's unique, why would you place the burden of remembering a ```username``` on the user? **BIG QUESTION: Do developers who need ```username``` actually have an ```email``` column?!**

For problems (2) and (3)... I guess we could:
(A) "magically" add the specified column to the "_my account_" and "_register_" forms. With the ```required```, ```string``` and ```unique``` validation. 
or
(B) change the "_my account_" and "_register_" forms to use ```username``` **instead of** ```email```, just like the login form does; we'd also disable the "_forgoten password_" functionality; 

I think solution B is better. But it makes a big assumption - that there is no required ```email``` column in the db.

### What to do?

It's tough for me to figure this out, since I haven't needed to switch to ```username```. Ever.  I can only rely on you guys (and intuition) on what to do. @kichetof , @iMokhles , @lloy0076 , @niladam what do you think?

Questions:
[Q1] **How many people need ```username``` in their admin panels? Please speak up.** Is this really a feature we should include in Backpack\Base, or should we just create a tutorial for how to achieve it?
[Q2] **Developers that need ```username``` - do you have an ```email``` column in your db?** This is very important, because we can figure out whether to use solution A or B above.

lloy0076:
Solution B.

Password reset could go along the lines of:

"Please enter your username. IF there is an associated e-mail, you will be sent one."

And behyind the scenes, either because of configuration, the fact there is an e-mail address or the correct coloured bantam was sacrificed, it would *just work*.

tabacitu:
@lloy0076 thank you for the feedback.

Ok, so. **I've fully implemented solution B for the ```username``` situation**, but I haven't changed the _password_reset_ procedure to use ```username``` instead of ```email```. Because (1) people who forget their ```password``` also forget their ```username``` (rarely their ```email``` though) and (2) It makes no sense to change the working procedure & error texts in all languages, for something that still uses email, after all. So I've just disabled the _password_reset_ functionality if the ```email``` column is not present on the users table.

This makes this implementation as good as it can be, from my point of view. It's Ready.

However. **I'm still not 100% that Backpack should support ```username``` out-of-the-box**. Keep in mind that I've worked quite a bit on this - the work has already been done. I'm just wondering if it's the right move to _merge_ it, from Backpack's perspective. The way I see it:

PROs:
- developers who need admin panels with ```username``` have a quick way of building that;

CONs:
- in order to make ```username``` an option, we've had to modify [quite a few stuff in Backpack](https://github.com/Laravel-Backpack/Base/pull/257/files); so we  now have a bunch of conditionals everywhere, for something that _might be_ a niche use case; I wouldn't quite say that _it's bloat_; but that's how good code turns into bloat - little by little; 
- if we provide out-of-the-box ```username``` support, we'll need to maintain ```username``` support; for ever and ever;
- all Backpack packages that deal with Users will be forced to support this new feature; current packages _and_ future packages; 
- as-is, PermissionManager will break for those who use ```username```; by default it allows admins to edit ```email```, not ```username```; we can provide support, but should we automagically allow them to edit the ```email``` column too, if present? nasty;
- ```username``` in authentication is something that Laravel _itself_ only provides partial support for; the docs instruct you to do one thing and it will "just work", but it doesn't; you still have to modify at least 3 more blade files and 2 more controllers to make Register and Forgotten Password to work; so it's a _custom feature_ for Laravel; not a _supported out-of-the-box feature_;

Again - this is mostly _my_ code, that I've spent a reasonable amount of time developing. So I want to merge it. I just don't know if it's a good idea. **I really need some help in deciding whether this is a feature to include, or not. What do you guys think?**

Then again, I might be overthinking this... It's been known to happen :-)

## Tutorial - to be included in the docs

How to login and register with ```username``` instead of ```email```, when/if this PR gets merged:
1. Create a ```username``` column in your users table and add it in ```$fillable``` on your ```User``` model. Best to do this with a migration.
2. Delete your ```email``` column and remove it from ```$fillable``` on your ```User``` model. Alternatively, just remove UNIQUE and NOT NULL from it. Best to do this with a migration.
3. Change your ```config/backpack/base.php``` config options:
```php
    // Username column for authentication
    // The Backpack default is the same as the Laravel default (email)
    // If you need to switch to username, you also need to create that column in your db
    'authentication_column' => 'username',
    'authentication_column_name' => 'Username',
```
That's it.

tabacitu:
Questions:
- Do you think this is bloat?
- Do you think this is a feature we should provide out-of-the-box?

Thank you!

lloy0076:
I honestly doubt it's a niche use case; but then again, if I could verify that simply by stating it, I'd be a pollster and not a tech "guru" 😮

src/app/Http/Controllers/Auth/ResetPasswordController.php
```diff
@@ -32,6 +32,10 @@ public function __construct()
     {
         $this->middleware('guest');
 
+        if (!backpack_users_have_email()) {
+            abort(403, trans('backpack::base.no_email_column'));
```
lloy0076: 403 feels wrong to me.

Personally I think it's one of the 500s or a view that states "Your super intelligent administrator must be contacted to do this."

src/helpers.php
```diff
@@ -27,13 +55,19 @@ function backpack_url($path = null)
      */
     function backpack_avatar_url($user)
     {
+        $placehold = 'https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0];
```
lloy0076: `placeholder` => english!

src/helpers.php
```diff
@@ -27,13 +55,19 @@ function backpack_url($path = null)
...
+
         switch (config('backpack.base.avatar_type')) {
             case 'gravatar':
-                return Gravatar::fallback('https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0])->get($user->email);
+                if (backpack_users_have_email()) {
+                    return Gravatar::fallback('https://placehold.it/160x160/00a65a/ffffff/&text='.$user->name[0])->get($user->email);
```
lloy0076: You need to escape the `text=` or eventually somehow something will go wrong.

https://groups.google.com/forum/#!topic/comp.mail.headers/MDZPvKXTb3w

src/app/Http/Controllers/Auth/ResetPasswordController.php
tabacitu: True. Changing it to 501 (not implemented).

src/helpers.php
tabacitu: :-)) true. Changing!

src/helpers.php
tabacitu: I wondered the same thing one day, so tried a bunch of email addresses to break it, but wasn't able to. Apparently Gravatar pipes this through a pretty solid method ```buildUrl()``` that takes care of everything :-)

tabacitu:

src/helpers.php
lloy0076: Relying on outside services for security is likely to get us in trouble.

EquiFax could purchase the Gravatar guys, for example.

tabacitu: That would be a treat :-) Gravatar _itself_ doesn't do the escaping. The Laravel package that makes the call to them does the escaping. If we escape it before that package escapes it, we'd just be escaping it twice.

kichetof:
Glad to see how you improved my PR!! :)
Everything works fine for me! And as you said, Backpack offer a great solution out of the box and if you want to customize it with any other login field, feel free :)

Backpack permission package should be edited to allowing `authentication_column` instead of default email value.
That the only last thing that I need to do to have Backpack setting up with Active Directory support with samAccountName field instead of email.

niladam:
@tabacitu I also agree that this shouldn't be the default authentication. Sticking to login in with email but **provide** the option to change that to username based feels great. It's an option worth having.

PR #267, opened 2018-03-26
Add commets for default middleware

cristianuibar:
We should better explain the default behavior here as after the upgrade to the latest version we did not know what the issue was, our routes were redirecting to /home by default.


tabacitu:
Thank you @cristianuibar ! We've actually made the ```web``` middleware a default, alongside ```bacpack_middleware()```. Launching without the ```web``` middleware was a huge blunder on my end - very sorry for the confusion and frustration I caused you.

Cheers!

PR #277, opened 2018-04-18
fix a fix: jQuery -> jquery (#189)

alashow:


alashow:
Okay, looks like this still won't work because of 095418ea6308cd74cbc7605654ea37181d73ac35.

alashow:
I will close this, and will create another branch for this.

PR #282, opened 2018-04-29
add topbar_content.blade.php to menu.blade.php

tswonke:
Now we can add items to the top menu as easy as to the sidebar.

I don't think that this file has to be published because I assume that this is not often used. But with this pull request we can just create a `topbar_content.blade.php` in our project and add items without overwriting `menu.blade.php`

### DIFF
--- a/src/resources/views/inc/topbar_content.blade.php
+++ b/src/resources/views/inc/topbar_content.blade.php
@@ -12,7 +12,7 @@
         </li>
         <li class="">
             <a href="{{ url(config('backpack.base.route_prefix', 'admin')) }}">
-                <i class="fa fa-list"></i> Dropdown Tiem
+                <i class="fa fa-list"></i> Dropdown Item
             </a>
         </li>
     </ul>


lloy0076:
It probably should be published with the other asset (sidebar) for consistency.

lloy0076:
And should it be optional via configuration:

1. Should it appear _at all_?
2. When should it appear?

...and so forth.

At the moment it looks as though this proposal believes it should appear when the _default_ guard responds "yes".

tswonke:
@lloy0076 
I understand your questions and one can think of diverse possible optimizations/configurations.
This was intended to be a very quick not-breaking change for a function that is probably not used very often.

btw: the used check (default guard) is the same as in `sidebar.blade.php`, or do I miss something?

AbbyJanke:
Everything looks good to me, saw you fixed the spelling 👍 I say its good to go, but i'm curious if we should add this to the "minimal" views that get published when running `backpack:base:install`.

tswonke:
Thank you @AbbyJanke !
I'm not sure about whether this should be published by default, what do you think @tabacitu ?

tswonke:
any change to get this merged @tabacitu ? :-)
what do you think about publishing by default?

tswonke:
Sorry, I will open another pull-request. I deleted this repo after being added to the team.

tswonke:
see #302

PR #289, opened 2018-05-26
CSS rule improvement + avoid CDN

promatik:
Minor change on css rule, `body.skin-purple` is faster than `body[class^='skin-purple']`, and it loads instantly
Formatted `backpack.base.css` (2kb savings)
Font awesome loaded from `vendor/adminlte` instead of cdn


promatik:
The css rule starts with is causing this weird glitch (at least on chrome):

image

When page is loading, the rules aren't instantly applied.
Only when document fully loads, the rules are applied.

image

tabacitu:
I think you’re totally right @promatik - thank you!

I merged your changes into the a different PR - see here https://github.com/Laravel-Backpack/Base/pull/332 so we can merge them into the ```dev``` branch (many changes since).

Thank you! Keep ‘em coming!

PR #293, opened 2018-06-04
[Merged] Fix password reset process; Completely separate Backpack authentication from scaffolded Laravel auth

tabacitu:
Password reset process has apparently stopped working since we introduced the custom auth guards functionality. I've tracked down the problem - seems to be because:
- the default password broker is ```user```;
- that password broker uses the authentication provider ```user```;
- that authentication provider uses the class ```App/User```, not the one in Backpack;

I suspect we did not detect this so far because we did not correctly upgrade our apps from 0.8 to 0.9. We did not delete the extra notification stuff from App/User. If you do, or you install Backpack on a new project, the password reset form crashes with an ugly error, and the user receives no reset email. So the process is broken.

### Solution A
In order to fix it, the only way I saw possible was to create:
- a custom auth provider: ```backpack_users```
- a custom password broker: ```backpack_users```

And use this password broker by default, according to the base config file. This _should_ be a non-breaking change, since we're still using the ```users``` auth provider, so the sessions are merged for login. But it's messy, since we have a ```backpack_users``` auth provider that we're not using for login. But if we do, the sessions will be separated by default. Tough choice.


tabacitu:
Maybe a simpler solution would be to instruct people to  use the BackpackUser model in their ```config/auth.php``` file, instead of ```App\User```:

```diff
    /*
    |--------------------------------------------------------------------------
    | User Providers
    |--------------------------------------------------------------------------
    |
    | All authentication drivers have a user provider. This defines how the
    | users are actually retrieved out of your database or other storage
    | mechanisms used by this application to persist your user's data.
    |
    | If you have multiple user tables or models you may configure multiple
    | sources which represent each model / table. These sources may then
    | be assigned to any extra authentication guards you have defined.
    |
    | Supported: "database", "eloquent"
    |
    */

    'providers' => [
        'users' => [
            'driver' => 'eloquent',
-           'model'  => App\User::class,
+           'model'  => Backpack\Base\app\Models\BackpackUser::class,
        ],

        // 'users' => [
        //     'driver' => 'database',
        //     'table' => 'users',
        // ],
    ],
```

tabacitu:
### Solution C

Maybe an even better solution would be to YES, have a ```backpack_users``` provider, and a ```backpack_users``` password broker, so that the authentication is completely separate, BUT those configurations to be actually inside the ```config/auth.php``` file. We could write them there automatically upon installation…

This would also allow developers to customize the provider and password broker if they need to.

### Solution D

A hybrid solution (A and C) would be to insert the provider and password broker at runtime (like A does), but only do so if there isn’t already a provider or broker with the same name. That would allow developers to customize the provider or password broker by inserting stuff into ```config/auth.php```, and but using the ```backpack_users``` name.

tabacitu:
This PR has been changed to apply **Solution D**, which seems the most sensible. **Backpack now ships with its own auth provider, password broker and guard**, all called _backpack_ (not _backpack_users_ as before).

By default, as specified in the ```config/backpack/base.php``` config file, Backpack uses this ```backpack``` provider, broker and guard. So **Backpack login is now completely separate from the scaffolded Laravel login**. The only thing they might have in common is the ```users``` table in the database.

tabacitu:
Merged into ```dev```. Will automatically be closed when this PR https://github.com/Laravel-Backpack/Base/pull/324 gets merged.

tabacitu:

PR #294, opened 2018-06-24
Changing Brazilian portuguese lang code to follow the international standard

claudsonm:
In most of the other packages and applications the lang code for Brazilian portuguese is "pt-BR", which follows the international standard adopted by [W3C](https://www.w3.org/International/questions/qa-choosing-language-tags) and documented by [IANA Language Subtag Registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry).

Using the [Language subtag lookup](https://r12a.github.io/app-subtags/), we can insert the lang code in the text input of the Check box and see if it is a valid one. This tool it's not case sensitive but because the standard says the region code should be, so I kept in that way.


claudsonm:
If approved, give me a heads up so I can replicate that to the other Backpack packages. In that way everything will be synced.

claudsonm:
In fact, this is something I tried to do but without success. The standard says it should be "pt-BR" but most of the times we put everything in lower case (in other things, like HTML lang tag). If there is any way to make it case insensitive I think will be perfect. But this is probably a Laravel thing and not about the package itself.

The bigger problem here is the underscore, which I've never ever seen before.

Em 25 de jun de 2018 8:56 AM, David Lloyd <notifications@github.com> escreveu:

@lloy0076 commented on this pull request.

I get why you want to do this and we probably should however this is likely to play havoc on those with case insensitive file systems 😢

tabacitu:
Hi @claudsonm  and @lloy0076 ,

Sorry for not getting to this sooner. At the time I thought this isn’t a big deal, but it kind of is. This inconsistent folder naming does lead to problems. Namely, if you use ```pt_br``` for your locale, then:
- Backpack will have the correct strings 
- other packages, that use ```pt-BR```, will NOT have the correct strings

I’ve dive deeper into this issue, and it’s a problem [symphony](https://github.com/symfony/symfony/issues/2468) has had, laravel has had and [jessengers/date has had](https://github.com/jenssegers/date/issues/145) (which we use for translating dates). And there’s no single solution. But **it does look like the dash-and-uppercase standard is more widely used in Laravel**, especially since [caouecs/Laravel-lang uses it](https://github.com/caouecs/Laravel-lang/tree/master/src), and that’s where most users get their extra language files. 

So it would probably be a good idea to migrate from ```pt_br``` to ```pt-BR```. And similarly for ```da_DK```, ```fr_CA```, etc. But this would be a breaking change for everyone using ```pt_br```, for example. So **I think, for now, it’s better if we just have folders for both ```pt_br``` and ```pt-BR```, with the same files inside. That way, we’re not introducing a breaking change for people using pt_br, but we provide a solution for people using pt-BR. And we beat a path to normality. What do you think @claudsonm ?**

tabacitu:
PS. Thank you for the well-documented PR and sorry again for not giving this importance sooner - I was wrong to do so.

claudsonm:
> So it would probably be a good idea to migrate from `pt_br` to `pt-BR`. And similarly for `da_DK`, `fr_CA`, etc. But this would be a breaking change for everyone using `pt_br`, for example. So **I think, for now, it’s better if we just have folders for both `pt_br` and `pt-BR`, with the same files inside. That way, we’re not introducing a breaking change for people using pt_br, but we provide a solution for people using pt-BR. And we beat a path to normality. What do you think @claudsonm ?**

@tabacitu I think it's ok for now to keep both folders, as long we provide some warning (as comments inside the files or at the documentation). Always remembering to get rid of the duplications in a new major version of the package.

tabacitu:
@claudsonm - yup, that’s the plan. I’ve created a separate PR for this, and will merge it into ```dev``` - https://github.com/Laravel-Backpack/Base/pull/333

Thanks again! Cheers!

PR #295, opened 2018-06-28
Removing abundant space in value attributes

pavoltanuska:
Both input fields have an extra space at the end, this causes adding a space at the end of the username after each save.

Funny thing is, this issue isn't visible on the second field, if it's `type=email` as I suspect that email field trims the extra spaces. When using `type=text` though, the issue is still there.


tabacitu:
Totally! Thank you @pavoltanuska , merged it!

PR #297, opened 2018-07-07
[4.0][Feature][Ready] Themes

AbbyJanke:
This allows for various theme options. While I suggest that we only support adminlte by default to avoid further update and such once we get a 'marketplace' type place setup themes should be included on there and could be downloaded etc. Takes less work from the core backpack team and spreads more to the community to help contribute, expand, etc. perhaps even get a category on themeforest and such.

Added two new helpers both return strings:

```
backpack_view($view, $package = 'backpack') { // }
backpack_theme($view, $package = 'backpack') { // }
```

- `backpack_view()`: This allows for us to get just the view, and we can define a package such as `crud` but defaults to backpack.
- `backpack_theme()`: Gets the theme to be used based on the new config option of `backpack.base.theme` (defaults to 'adminlte')


AbbyJanke:
CRUD PR: https://github.com/Laravel-Backpack/CRUD/pull/1513

tabacitu:
Hi @AbbyJanke ,

Thanks a lot for the PR. You know that just like you, I've wanted themes for a while now :-) Can't wait to be able to use Backpack with tabler :-D

Looks good to me so far, thank you!

AbbyJanke:
:) an ultimate goal I would like to see is be able to upload them via the admin panel with a package “theme manager” enable/disable them per user etc. debating if uploading like a .zip or entering a custom packagist site for themes.

shirshak55:
@tabacitu  At this point I don't recommend tabler :) If we want to use tabler we should probably maintain it ourself.

Probably these themes are better as of now

https://github.com/DesignRevision/shards-vue
https://github.com/puikinsh/Adminator-admin-dashboard

Do you have any plan for vue?

tabacitu:
@shirshak55 I think Shards looks & works great on front-end. Really like it, thank you, didn't know about it. Don't know how complete it would be for back-end (admin panels) though. I personally find adminator good enough, though I think its spacing & padding is a little off / inconsistent. Easily fixable, though.

Our theming system should allow _anyone_ to create a package so that _everyone_ can use Shards. Or Adminator. I guess in the end it shouldn't matter what I think :-) Backpack should come with a default theme which would "_do the job_", but others themes would be available for each of these good options.

Vue, imho, should not be a requirement to use Backpack. That would be too opinionated. So we don't plan on using vue-router. But I don't see why anyone could not use Vue in parts of a Backpack-powered admin panel - for a field, for a dashboard, for other custom pages. Just like it would be ok to use Angular, or anything else. That's my take on this one.

shirshak55:
@tabacitu we are already opinionated because we use jQuery isn't it? Due to that we cannot easily move to Vue or React or Angular. There are already too many vue fans and react fans which we cannot ignore also. So I think we can do it.

iMokhles:
Nice start @AbbyJanke 
but i think it could be easier than that as we have this [PanelTraits/Views.php](https://github.com/Laravel-Backpack/CRUD/blob/master/src/PanelTraits/Views.php) trait,
we need just to include the new layout using a simple method we can call it 
`backpack_layout('layouts.admin_layout')`
then inside 
`XXXXCrudController.php`
we can now set new views using the `PanelTraits/Views.php` trait.
yes i know that there are fields and columns needs to be changed.
so actually the only way to do it through creating new views, because current fields and columns aren't flexible enough to do that even if you can set custom views for fields it's not that flexible.


hope we can find away which based on layout change.

tabacitu:
Merged into the ```v4``` brach. Can't wait :-)
Thanks again for submitting this @AbbyJanke !

shirshak55:
wow thanks for this release :)

tabacitu:
@shirshak55 you're welcome :-) But please note it's NOT released yet - it's still alpha. Base is pretty much ready, but we're still merging BIG breaking changes in CRUD. I hope the beta will be ready at the end of the week. Can't wait to get feedback :-)

PR #300, opened 2018-07-16
[Feature][0.9][Ready] Added version command to output PHP, MySQL and Backpack versions

tabacitu:
Added a new command that should help us debug stuff. If we merge this we'll be able to ask people to run ```php artisan backpack:base:version``` and copy-paste the output, which will include:
- the PHP version
- the version of all backpack & laravel packages
- the MySQL version

Credits go to @OwenMelbz and [his comment here](https://github.com/Laravel-Backpack/CRUD/issues/1539#issuecomment-404504641).

Output example:
screen shot 2018-07-16 at 19 24 54

[Q1] I wonder what happens if MySQL isn't installed?!

### DIFF
--- a/src/BaseServiceProvider.php
+++ b/src/BaseServiceProvider.php
@@ -12,6 +12,7 @@ class BaseServiceProvider extends ServiceProvider
         \Backpack\Base\app\Console\Commands\Install::class,
         \Backpack\Base\app\Console\Commands\AddSidebarContent::class,
         \Backpack\Base\app\Console\Commands\AddCustomRouteContent::class,
+        \Backpack\Base\app\Console\Commands\Version::class,
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


OwenMelbz:
You should be apple to wrap the ands in ors as well sooo something like.............

`x && (mysql --version || "")` so it might still say an error, but wont fatal out?

tabacitu:
@OwenMelbz you're right, that should work. 

Ended up splitting them into separate processes, since a the output desperately needed a few headers too. So 2 birds with one stone. Now output is prettier and should fail nice too:

screen shot 2018-07-16 at 20 28 02

OwenMelbz:
Looking good :)

If you wanted to go further you could extend it do have like :base:dump or something that creates a text file with it all in?

You could also like PHP modules, all packages etc

tabacitu:
Hmm... PHP modules would be nice, but I don't think they're really necessary. If something fails because of that, it'll probably throw an ugly obvious error.

Dump to file would also be nice, but... not necessary either, I think. You don't need to _store_ that information, just copy-paste.

PR #302, opened 2018-08-01
[Merged] Feature: edit topbar content

tswonke:
This adds two files (`inc.topbar_left_content.blade.php` and `inc.topbar_left_content.blade.php` for easy adding of topbar menu elements.

Sorry for the multiple "changes" in `menu.blade.php` most of it is due to code styling.

Please see #282 - I had to open a new pull request for this.

@tabacitu do you think we should publish these files by default? I'm not sure about it.


tabacitu:
Thanks for the PR @tswonke . I'm not sure this is a necessary change either. It would TOTALLY make sense if we give developers the ability to generate sidebar items into these two files too, instead of the sidebar. So _when generating a Test CRUD_, a menu item could be added to ```topbar_left_content``` or ```topbar_right_content``` instead of ```sidebar_content```. Maybe depending on a config option.

BUT. I'm not sure that's something anybody wants/needs. I myself have tried having links to CRUDs in the top bar, but since there are usually 8-10 CRUDs, it quickly becomes crowded. So I've either placed them back in the sidebar, or under a dropdown on top. And we can't know by default what dropdown developers have.

Since people can already edit the contents of the top menu now, by editing the existing file, I think this _nice_ (definitely more consistent with how the sidebar menu works), but not needed. **Let's see what others think**, and we'll take it from there. In my opinon, this is a nice consistency change, but we shouldn't consider it a feature-bringing PR. I don't think we should give developers the ability to choose where to generate their CRUD links.

Thanks again. Cheers!

PS:
- Please don't hold it against me if we end up not merging this. I myself have a number of PRs I proposed that we did not merge.
- If we do end up merging this, i think we should move the auth buttons from ```menu``` to ```topbar_right_content```. This way developers won't ever need to change ```menu```, only the content blade files.
- Since this PR only adds blade files, it doesn't remove anything, I think we can consider this a non-breaking change.

tswonke:
Thanks for the feedback @tabacitu 

I use the topbar for alert dropdowns (s. AdminLTE Examples) in a project and just thought it might be interesting to keep the `menu.blade.php` clean. But I understand your concerns and I'm not sure whether this PR is useful for "all" of us.

tabacitu:

PR #305, opened 2018-08-03
Fix backpack_url helper

mikemand:
Detection of a leading slash was buggy. If I did something like: `{{ backpack_url('/settings') }}` I would end up with a URL like: `/admin//settings`.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -11,7 +11,7 @@ if (!function_exists('backpack_url')) {
      */
     function backpack_url($path = null)
     {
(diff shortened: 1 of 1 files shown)


tabacitu:
Totally agree! Thank you @mikemand - that's definitely the way it should have been from the start. Since I don't think people were _intentionally_ using it that way, let's consider this a non-breaking change. I've merged it into ```master```, it'll be available with a ```composer update``` shortly, after I bundle it with a few more non-breaking changes.

Thanks again for the PR. Cheers!

PR #310, opened 2018-08-30
Create user command

AbbyJanke:
Allow the user to create a new user via console:

```
backpack:user --name=Abby --email=abby@example.com --password=mypassword
```

Features:
- If any of the options (name, email, password) are left out it will prompt for input.
- `--encrypt` option, if it should auto encrypt or not (useful for using the command within controllers to create a new user.
- option shortcuts: `N` for name, `E` for email, and `P` for password.

### DIFF
--- a/src/app/Console/Commands/CreateUser.php
+++ b/src/app/Console/Commands/CreateUser.php
@@ -45,8 +45,8 @@ class CreateUser extends Command
             $password = $this->secret('Password');
         }
 
-        if(! $this->option('encrypt')) {
-          $password = bcrypt($password);
+        if (!$this->option('encrypt')) {
+            $password = bcrypt($password);
         }
 
         $auth = config('backpack.base.user_model_fqn', 'App\User');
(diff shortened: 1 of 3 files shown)


src/app/Console/Commands/CreateUser.php
```diff
@@ -44,11 +45,15 @@ public function handle()
             $password = $this->secret('Password');
         }
 
+        if(! $this->option('encrypt')) {
+          $password = bcrypt($password);
```
lloy0076: Oh how I hate that helper - wouldn't it be better to use the `Hash` facade? Especially since Argon2 is now supported...

AbbyJanke: My reasoning for using the bcrypt helper is that by default that is what vanilla Laravel and Backpack uses by default within the `RegisterController`. Was trying to stick with what is currently used.

tabacitu:
This is excellent! Thank you @AbbyJanke ! I made just a few modifications which I think make it more intuitive:
- I think the password SHOULD be encrypted by default (so you have to specify ```--encrypt=false``` to NOT encrypt it);
- I renamed the command from ```backpack:user``` to ```backpack:base:user```, for consistency; all command so far follow this convention, so even though it's a little bit uncomfortable to think about it that way (as part of base), I think we should keep the convention;

Thank you again. Cheers!

PR #315, opened 2018-09-11
add robots meta tag and config parameter

tswonke:
This PR adds a robots meta tag to `layout.blade.php` and a config parameter to adjust the content. Intention is to prevent indexing and following links in admin backends.

### DIFF
--- a/src/config/backpack/base.php
+++ b/src/config/backpack/base.php
@@ -37,7 +37,7 @@ return [
     'default_datetime_format' => 'j F Y H:i',
 
     // Content of the HTML meta robots tag to prevent indexing and link following
-    'robots_content' => 'noindex, nofollow',
+    'meta_robots_content' => 'noindex, nofollow',
 
     /*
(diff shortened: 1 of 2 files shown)


OliverZiegler:
Would add if statement to prevent rendering if nothing was set... 🤔 

```php
@if(!empty(config('backpack.base.robots_content')))
   <meta ....>
@endif
```

tswonke:
Thanks @OliverZiegler 

My thought was, that "noindex, nofollow" could/should(?) be kind of a standard. That's way I used a default value...

Opinions? @tabacitu ?

tabacitu:
Great addition, thank you @tswonke and @OliverZiegler . Totally agree with both of you:
- adding robots meta ```noindex,nofollow``` as default makes total sense;
- if you don't want it, you should be able to just leave it empty - then it won't output the meta at all;

Made the changes (and changed the config to ```meta_robots_content``` to be more specific). Will be tagged later today.

Cheers!

PR #321, opened 2018-10-09
Fix active links when have GET parameters in URL

savicdragan15:
Fix active links when have GET parameters in URL.


tabacitu:
Thank you for the PR @savicdragan15 ! Looks nice and clean.

**Can you please give me an example?** An URL where the wrong link was active? I can't seem to find a use case myself, even if I'm using the entity name as GET parameter, value, etc.

Thanks, cheers!

PR #323, opened 2018-10-16
[Merged] BackpackUser model picks up its parent's relationships

tabacitu:
By default, we use BackpackUser model everywhere, and ```BackpackUser extends User```. But the way Eloquent works, that extension DOES NOT include relationships. So when you have Roles, Permissions or other relationships on the User model, the BackpackUser object WILL NOT have those relationships.

This makes it more difficult to use packages like PermissionManager (see issues https://github.com/Laravel-Backpack/PermissionManager/issues/169, https://github.com/Laravel-Backpack/PermissionManager/issues/166), but also with home-grown relationships. You'd expect if you add a relationship to the User model, that BackpackUser would also pick it up. Right now it DOES NOT. And the helper we recommend using for getting the current user, ```backpack_user()```, returns a BackpackUser model by default.

Apparently forcing Eloquent to pick up its parent's relationships isn't that easy, but there's [a package by Tighten that does the job](https://github.com/tightenco/parental). They say it's alpha, but in my experience it works really well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 7.0
   - 7.1
   - 7.2
   - nightly
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ Updates should follow the [Keep a CHANGELOG](http://keepachangelog.com/) princip
 
 -------
 
+## [0.9.10] - 2018-10-15
+
+### Fixed
+- loading the custom views folder is now done only if it exists, this way fixing conflicts with the ```php artisan view:cache``` command;
+
+
 ## [0.9.9] - 2018-09-20
 
 ### Added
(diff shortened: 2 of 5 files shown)


tabacitu:
**Note:** this package does not support Laravel 5.5.

So it can only be merged in a bumped version of Backpack\Base, where we drop support for Laravel 5.5 - which was due anyway.

tabacitu:
Merged into the ```dev``` branch. Will be closed automatically when https://github.com/Laravel-Backpack/Base/pull/324 gets merged.

tabacitu:

mpixelz:
looking forward for this..

i just started playing with different guards for admin and users.. but now cant use @can() functions in the admin section.. as they gets ignored altogether.

tabacitu:
Yes - tha plan is to launch today or tomorrow 🤞 That’s what I said last week too, but I kept cramming more and more features and bug fixes into this update - it’s going to be a big one 😄

PR #326, opened 2018-10-25
Update to use primary key instead of ID

Lzxe92:
Rule::unique($user->getTable())->ignore($user->getKey()) 
By default uses the user table of idColumn = id. Added additional attribute of getKeyName to use the primary key defined in the user table.


tabacitu:
Thank you @Lzxe92 ! You’re totally right - thank you for the PR. Just merged and tagged it, so you can do a ```composer update``` to get the fix yourself.

Cheers! Keep ‘em coming :-)

PR #328, opened 2018-10-30
[Merged] Error views are now consistent with main adminlte design

tabacitu:
- changes error views to use a layout, so it’s easier to customize the design for all error pages
- changes default error views design background, font and color to match AdminLTE

Before:
screenshot 2018-10-30 at 10 01 36

After:
screenshot 2018-10-30 at 09 58 09

### DIFF
--- a/src/resources/error_views/400.blade.php
+++ b/src/resources/error_views/400.blade.php
@@ -1,60 +1,16 @@
-<html>
-  <head>
-    <title>{{ config('backpack.base.project_name') }} Error 400</title>
+@extends('errors.layout')
 
-    <link href='//fonts.googleapis.com/css?family=Lato:100' rel='stylesheet' type='text/css'>
+@section('error_number')
+  400
+@endsection
 
-    <style>
-      body {
-        margin: 0;
-        padding: 0;
-        width: 100%;
-        height: 100%;
-        color: #B0BEC5;
-        display: table;
-        font-weight: 100;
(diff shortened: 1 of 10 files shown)


tabacitu:
Since these views are published upon installation, existing installs will not get this change, unless they delete the ```resources/views/errors``` folder and re-publish it.

However, since this is such a trivial change and nobody will probably mind it, I say let’s merge it as a non-breaking change.

tabacitu:
Merged into ```dev```. Will get closed when the upgrade branch https://github.com/Laravel-Backpack/Base/pull/324 gets merged

tabacitu:
Note: I ended up making some more changes to this:
- the error layout view now uses the guest layout that we use for authentication, by default
- if the admin is logged in and on an admin page (which we determine using the admin prefix), the error layout view now uses the backpack layout

What this does is that **if the admin is logged in and runs into and error, he will still view the sidebar**. So he has a quick way out. I found it very annoying to run into something and only have "back" and "go home" buttons - sometimes I say "fuck it, I want to go somewhere else then" and I can't because there's no sidemenu. And then it's two clicks away. Who likes two clicks away? :-) 

Examples (logged in and logged out):
screenshot 2018-11-07 at 10 26 18
screenshot 2018-11-07 at 10 27 44

tabacitu:
(the above has already been pushed to ```dev```)

PR #330, opened 2018-10-30
Guest layout, prettier auth views

tabacitu:
Applied all changes @iMokhles has done in this PR https://github.com/Laravel-Backpack/Base/pull/233 - excellent idea and great implementation, didn’t get to merge it because there were too many merge conflicts (my fault). @iMokhles - I did not use AdminLTE’s design for the auth pages in the end - I think our is actually better now :-)

Changes:
- added a guest_layout, which has no sidebar and no menu
- made footer transparent
- polished all auth pages so that they’re simpler and have no sidebar
- password reset link now prepopulates the email field
- made it obvious that reset password is a two-step process

Screenshots:
screenshot 2018-10-30 at 15 30 17
screenshot 2018-10-30 at 15 30 25
screenshot 2018-10-30 at 15 30 37
screenshot 2018-10-30 at 15 35 05


iMokhles:
great but i guess the tab option should be for signup/signin only.

tabacitu:
Hmm.. what do you mean @iMokhles ? If you mean the “step 1, step 2” tabs, they’re only for the “reset password” procedure, which has always been a two-step process, but has never been show as such.

iMokhles:
yes the one for reset password i guess it should be for the signup and signin steps only, i don't see a need for it under reset pages, because use won't be able to use the 2nd step to reset the password unless he click on the reset link received by email.

tabacitu:
… I’m afraid I don’t understand what you mean. 

I think:
- the tabs DO make sense on reset’s 1st and 2nd step
- the tabs DON’T make sense on login/register, since they’re one-step processes

I’m reading that your opinion is the other way around, maybe I’m misreading… Could you please clarify?

Also, they look even better now :-)

screenshot 2018-10-31 at 10 44 26

iMokhles:
yes looks great now :) but :P 

Steps concept works like that 

we fill the information in the 1st step to enable the 2nd one, 
but here if user request a new password within the 1st step, he won't be able to fill the 2nd step without going back to the reset email and click on the reset link which contains the reset token, so having tabs for reset pages won't be necessary.

and it will raise the amount of support issues for the developers, users will ask fo example why i wasn't able to use the 2nd step even i requested a new password within the 1st step ) most of them won't understand that they need to go back to the email and click on the reset link, and will try direct the 2nd step.

hope it's clear now ^_^

tabacitu:
Aaaah. Ok, I understand now :-)

Hmm... it might be confusing for people who expect to be able to get to the next step immediately, true, but I’ve seen this multi-step process with back-and-forth in multiple places. So it’s not uncommon, and I think it’s something that can be easily fixed with text. After you punch in your email it should read “_An email has been sent to your email address. Please click the link in the email to proceed to step 2._”

That would fix it, right?

PR #334, opened 2018-11-08
[1.0.0][Feature][Ready] Publish backpack user model & middleware upon installation

tabacitu:
In response to the discussion here: https://github.com/eduardoarandah/UserManager/issues/5

This adds two commands:
- ```php artisan backpack:base:publish-user-model```
- ```php artisan backpack:base:publish-middleware```

These commands are run BY DEFAULT when someone installs Backpack\Base.

They will publish those files inside the users' App folder, so that he can modify them at will.


tabacitu:
Since we're at it, we should _probably_ also publish the DashboardController. This is something that gets overwritten a lot.

Todo:
- [ ] add ```php artisan backpack:base:publish-dashboard-controller``` command

tabacitu:
No. Screw that. We don't publish the dashboard. We don't publish the Auth controllers. That would be too much :-)

What we _can_ do is instruct people in how to modify their ```dashboard.blade.php``` view, and use [view composers](https://laravel.com/docs/5.7/views#view-composers) to push info into that view.

tabacitu:
Done - added [better dashboard overwrite docs](https://backpackforlaravel.com/docs/3.5/base-how-to#customize-the-dashboard).

tabacitu:
Merged into https://github.com/Laravel-Backpack/Base/pull/324

Let's DO THIS!!!

PR #336, opened 2018-11-17
MakeCrud command

MarcBrillault:
I recently created this command to automate the crud creation in my projets, I hope it'll be useful for others too.

### DIFF
--- a/src/app/Console/Commands/MakeCrud.php
+++ b/src/app/Console/Commands/MakeCrud.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Artisan;
+use Illuminate\Console\Command;
+
+class MakeCrud extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'make:crud {name}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Creates a new backpack crud';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        $argument  = str_singular($this->argument('name'));
+        $plural    = title_case(str_plural($argument));
+        $snakeName = snake_case($argument);
+        $camelName = camel_case($argument);
+
+        $routeArgument   = sprintf("CRUD::resource('%s', '%sCrudController');", $snakeName, $camelName);
+        $sidebarArgument = sprintf(
+            "<li><a href='{{ backpack_url('%s') }}'><i class='fa fa-question-circle'></i> <span>%s</span></a></li>",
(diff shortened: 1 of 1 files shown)


tabacitu:
Totally! Thank you @MarcBrillault ! Will check this out and merge it ASAP - before the 3.5 launch this week.

tabacitu:
Hmm actually on second thought, I think this:
- would be better put inside the CRUD project (since it generates a CRUD);
- would be a good starting point for using https://github.com/webfactor/laravel-generators instead of Generators; could provide people an _alternative_ way to generate CRUDs, so that we ease out of using our own generators, that way we no longer recommend generating ```setFromDb()``` CRUDs;

This would also:
- allow us to gradually improve this command, and provide wizard-like functionality, like mentioned here https://github.com/Laravel-Backpack/Generators/issues/5#issuecomment-246165301 (not from day 1 though)

MarcBrillault:
Ok, I LOVE the command-line wizard proposition. I'll keep my implementation on my websites and wait for this to be done (Is there a chance to see it in 3.5 ?)

tabacitu:
@MarcBrillault definitely in 3.5. But it's a non-breaking change, so we'll push it after launch. I had to launch 3.5 already - I kept postponing it every week, to cram more features in it, but 2 months of postponing is enough, I think :-)

I also postponed requiring @tswonke & @OliverZiegler 's [generators package](https://github.com/webfactor/laravel-generators) in CRUD for after the 3.5 launch, for this reason. I'll take this opportunity to apologise to them too. But we had to _ship already_ :-) and since this will be a non-breaking change, it's better this way, I think.

I think it's best that we:
- polish this ```php artisan make:crud``` command to have a wizard;
- the ```make:crud``` wizard would use their much better generators;
- this ```make:crud``` wizard would become the default way of generating CRUDs;
- the old way of generating CRUDs (with ```setFromDb()```) would still be available somewhere deep inside the docs, but it wouldn't be the _recommended_ way;

I think we can finish this in 1-3 weeks, and have it ready for the December newsletter. It will be a non-breaking change, since we're not removing the old generators, but a GREAT improvement. This way:
- new users will not even know about the old way, since it's deep in the docs;
- old users will definitely prefer using the new ```make:crud```, since it doesn't require you to know any syntax, doesn't need any documentation, and is easy to remember;
- old users who will prefer the old way (though I don't imagine why) will still be able to have their old way;

tabacitu:
Update: 

Now, in 2020, we have `php artisan backpack:crud ModelName` which does exactly this, and it outputs:
- the model (or adds the CrudTrait to an existing model)
- the CrudController
- the CrudRequest
- the sidebar item

For more info, check out https://github.com/laravel-backpack/generators#usage and https://backpackforlaravel.com/docs/4.1/crud-tutorial

Closing this, since it's no longer needed. Thanks again @MarcBrillault & @tswonke for the help!

PS. Soon, thanks to Blueprint + DevTools (WIP), we'll be able to also:
- generate and run the migration
- generate factory
- generate seeder

Here's a sneak peek of Backpack\DevTools v0.1:

PR #337, opened 2018-11-18
add config to fix layout

tswonke:
Using `fixed` for this config option the layout will be fixed (sidebar and top navbar). So one can scroll the content, but the to bars stay visible

### DIFF
--- a/src/resources/views/layout.blade.php
+++ b/src/resources/views/layout.blade.php
@@ -44,7 +44,7 @@
     <script src="https://oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
     <![endif]-->
 </head>
-<body class="hold-transition {{ config('backpack.base.skin') }} sidebar-mini">
+<body class="hold-transition {{ config('backpack.base.skin') }} sidebar-mini {{ config('backpack.base.layout', '') }}">
 	<script type="text/javascript">
 		/* Recover sidebar state */
 		(function () {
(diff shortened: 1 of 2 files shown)


src/resources/views/layout.blade.php
```diff
@@ -44,7 +44,7 @@
     <script src="https://oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
     <![endif]-->
 </head>
-<body class="hold-transition {{ config('backpack.base.skin') }} sidebar-mini">
+<body class="hold-transition {{ config('backpack.base.skin') }} sidebar-mini {{ config('backpack.base.layout', '') }}">
```
tabacitu: Hmm I see there are _two_ configs being thrown out on the same line, in the same ```class``` attribute here. So I'm wondering - couldn't people achieve the same thing by changing the ```skin``` to ```skin-purple fixed```?

If so, maybe a better solution would be to:
- (A) do nothing, let them use ```skin``` instead of ```layout```
- (B) rename the ```skin``` config to ```body_classes``` and have it default to ```hold-transition skin-purple sidebar-mini```

What do you think @tswonke ?

tswonke: Lol, you're right... it seems that I was a blockhead 😂 

`body_classes` might be a good idea, but I wouldn't change it (for now) - maybe in 4.0 with some other customization configs...

PR #339, opened 2018-11-20
Added better explanation to encrypt option within CreateUser command

iMokhles:
Is the user's password already encrypted isn't the correct phrase for encrypt option concept, because if it's true it will encrypt the password using bcrypt so we shouldn't ask developers is this password encrypted or not this will confuse them because.


tabacitu:
Totally! Thank you @iMokhles !

PR #340, opened 2018-11-24
LV language translation updated

tomsb:
Also removed duplicate key from EN language file.


tabacitu:
Great! Thank you @tomsb , merged!

Issue #342, opened 2018-11-26
Hide developer link if not configured


johnpuddephatt:
Should the developer link only be shown if both developer name and link are set in config? Otherwise "Handcrafted by" can appear on its own, or an anchor with an empty href.

Happy to submit a PR.

tabacitu:
Merged, thanks for the PR!

PR #348, opened 2018-12-06
Include developer link only if set in config

johnpuddephatt:
In response to #342


tabacitu:
Great, merged, thank you @johnpuddephatt ! :-) Please see the message above too.

PR #359, opened 2019-02-09
[4.0][Ready] Parental upgrade to 0.6 (the single table inheritance package)

ziming:
0.6 of https://github.com/tightenco/parental has been released with several big bug fixes and enhancements to the single table inheritance implementation

This is a breaking change as the name of the Trait has changed in 0.6.

From HasParentModel to HasParent

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "prologue/alerts": "^0.4.1",
         "jenssegers/date": "^3.2",
         "creativeorange/gravatar": "~1.0",
-        "tightenco/parental": "0.5"
+        "tightenco/parental": "0.6"
     },
     "require-dev": {
         "phpunit/phpunit": "~6.0",
--- a/src/app/Models/BackpackUser.php
+++ b/src/app/Models/BackpackUser.php
@@ -4,11 +4,11 @@ namespace Backpack\Base\app\Models;
 
 use App\User;
 use Backpack\Base\app\Notifications\ResetPasswordNotification as ResetPasswordNotification;
-use Tightenco\Parental\HasParentModel;
+use Tightenco\Parental\HasParent;
 
 class BackpackUser extends User
 {
-    use HasParentModel;
+    use HasParent;
 
     protected $table = 'users';
 


AbbyJanke:
I recommend we save this till CRUD 4.0 comes out since it is a breaking change.

ziming:
In the meantime do you know of a way to force install a higher version of a package without composer throwing error? Say if someone need the stuffs (fixes and features) in parental 0.6 now

ziming:
Ok, I think I found out how for anyone interested. In your composer.json require section, do this:

"tightenco/parental": "0.6 as 0.5"

Then change your trait from HasParentModel to HasParent in BackpackUser model class

ziming:
Actually, how about removing parental requirement?

I see that BackpackUser has no relationship methods after all. A $table = 'users' on the class would have been good enough

ziming:
When is 4.0 coming? :)

tabacitu:
Totally agree with @AbbyJanke . Let's hold this off until 4.0 - shame, the breaking change was so minor...

@ziming thanks a lot for the PR! Very appreciated.

ziming:
Sure in the meantime just to leave a note for others.

You can do this in composer for the time being. Works fine for me so far

```
"tightenco/parental": "0.6 as 0.5"
```

tabacitu:
@ziming in the Base 1.1 update that I'm working on, **I've decided to remove the tightenco/parental dependency altogether**, and include the trait we're actually using as-in in Backpack/Base. This should:
- allow us to make sure they don't push any breaking changes as non-breaking (it's a 0.6 version after all, who know)
- allow us to make the upgrade process smoother in case they introduce any feature we'd like
- allow developers to use our trait, or theirs, if they decide to install the package separately
- remove a dependency, which is always good

Let me know what you think, the actual update is here: https://github.com/Laravel-Backpack/Base/pull/362/files#diff-488d31054d94cae3070e8fdb0a2efccd but there's not much to look at in this regard - I copy-pasted the Trait we're using from their package, and gave it a different name. The upgrade guide will instruct people to use this trait, instead of the one in the package.

Cheers!

PR #362, opened 2019-02-27
[WIP] Laravel 5.8 support

tabacitu:
This branch and PR will provide support for Laravel 5.8.

[TLDR here](https://github.com/Laravel-Backpack/Base/pull/362#issuecomment-468563782).

Todo:
- [x] require laravel 5.8 in composer.json
- [x] remove tightenco/parental dependency - since it does not yet support Laravel 5.8, I got the change to take another look at it; it look like we're only using it in conjunction with PermissionManager, so it makes no sense to force this dependency in Backpack/Base; We should do so in PermissionManager, and instruct users to add this to their BackpackUser model upon installation;
- [x] update or remove jenssegers/date dependency;
- [x] upgrade to nesbot/carbon v2; Note: reverted to carbon v1;
- [x] (maybe) bump adminlte dependency Note: was impossible; adminlte respository is now broken for most users;
- [x] upgrade bootstrap dependency to have recent fix to XSS vulnerability

Official packages TODOs:
- [x] Confirm / add L5.8 support to Backpack\PermissionManager;
- [x] Confirm / add L5.8 support to Backpack\Generators;
- [x] Confirm / add L5.8 support to Backpack\Settings;
- [x] Confirm / add L5.8 support to Backpack\PageManager;
- [ ] Confirm / add L5.8 support to Backpack\BackupManager;
- [x] Confirm / add L5.8 support to Backpack\LogManager;
- [x] Confirm / add L5.8 support to Backpack\LangFileManager;
- [x] Confirm / add L5.8 support to Backpack\MenuCRUD;
- [x] Confirm / add L5.8 support to Backpack\NewsCRUD;
- [ ] upgrade Demo to Laravel 5.8;
- [ ] write and publish upgrade guide;
- [ ] send newsletter;

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,30 @@ Updates should follow the [Keep a CHANGELOG](http://keepachangelog.com/) princip
 
 -------
 
+## [1.1.0] - 2019-03-01
+
+### Added
+- Laravel 5.8 support;
+
+### Fixed
+- Bootstrap was forced to upgrade to 3.4.1 by using a CDN, so that everyone gets the new XSS security vulnerability fix, and the upgrade process is smooth;
+
+### Removed
+- ```Tightenco\Parental``` dependency; the trait we were using is now included as ```Backpack\Base\app\Models\Traits\InheritsRelationsFromParentModel;```;
+
+**Upgrade guide:**
+- Upgrade to Laravel 5.8 or do ```composer require backpack/base:"1.1.*"
+- in your ```App\Models\BackpackUser``` instead of ```Tightenco\Parental\HasParent```, please use ```Backpack\Base\app\Models\Traits\InheritsRelationsFromParentModel```; [here's the diff](https://github.com/Laravel-Backpack/Base/pull/362/files#diff-f075b83ebb2b1ef3ba84dec14b395607);
+- if you've overwritten ```inc/head.blade.php``` or ```inc/scripts.blade.php```, please make sure you [use the newest version of Bootstrap](https://github.com/Laravel-Backpack/Base/pull/362/files#diff-96ac3ea4d0cb85053acf44e3772eb5f1); they've fixed a security vulnerability (XSS);
+
+
+----
+
+## [1.0.4] - 2018-12-10
+
+### Fixed
+- Portuguese translation;
+
 
 ## [1.0.4] - 2018-12-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,12 +48,6 @@ Updates should follow the [Keep a CHANGELOG](http://keepachangelog.com/) princip
 - Portuguese translation;
 
 
-## [1.0.4] - 2018-12-10
-
-### Fixed
-- Portuguese translation;
-
-
 ## [1.0.3] - 2018-12-06
 
 ### Added
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,7 @@
         "almasaeed2010/adminlte" : "~2.4",
         "laravel/framework": "5.6.*|5.7.*|5.8.*",
         "prologue/alerts": "^0.4.1",
-        "creativeorange/gravatar": "~1.0",
-        "tightenco/parental": "^0.6.1"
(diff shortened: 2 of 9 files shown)


AbbyJanke:
oof I've been soo busy with mobile apps I didn't even realize 5.8 was out. :( i miss easy php laravel dev..

tabacitu:
Haha @AbbyJanke mobile isn't that bad either :-) But yeah, you'd be welcome back.

**Upgrade slightly postponed because of dependencies:** 

1. AdminLTE tag - Upgrade to AdminLTE 2.4.9 (and re-publish) would also update to the latest Bootstrap 3, which would include the fix to the XSS vulnerability. Right now 2.4.9 does not seem properly configured on Packagist, though, and can't be installed with ```composer update```. I've [asked the creator to do that](https://github.com/almasaeed2010/AdminLTE/issues/2038#issuecomment-467931168), now we wait.

2. [tightenco/parental](https://github.com/tightenco/parental) to support Laravel 5.8. Right now this branch works, by eliminating the parental dependency. Because you don't necessarily need it. Not if you don't have Groups and Permissions for your users. But in order to have a working Backpack + PermissionManager we need to:
- [ ] 2.1. remove parental from Base;
- [ ] 2.2. add parental dependency to PermissionManager;
- [ ] 2.3. add installation step to PermissionManager - people should add parental trait on their User model;
- [ ] 2.4. add upgrade step in upgrade guide, because nonetheless the name and location of the parental trait has changed, so people need to edit the User model accordingly;

We can't do 2.2 because parental doesn't support Laravel 5.8 yet... We have to wait for them to add support.

tabacitu:
**Notes:**
- I've decided that Base 1.1 use Carbon v1, instead of upgrading to Carbon v2; this way, Backpack/CRUD will suffer zero changes, and support Laravel 5.6-5.8; We should upgrade to Carbon v2 and drop jenssegers/date in Base 1.2;
- I've decide to remove the tightenco/parental dependency, and include the one trait we need inside Base itself; since it's not a stable package I think that's safer - right now the only breaking change we have, and the reason we can't call this "1.0.x" is that they changed file names; that might happen again in the future; we might decide to install it properly when it reaches 1.0;

**Upgrade guide:**
- Upgrade to Laravel 5.8
- in your ```App\Models\BackpackUser``` instead of ```Tightenco\Parental\HasParent```, please use ```Backpack\Base\app\Models\Traits\InheritsRelationsFromParentModel```; [here's the diff](https://github.com/Laravel-Backpack/Base/pull/362/files#diff-f075b83ebb2b1ef3ba84dec14b395607);
- if you've overwritten ```inc/head.blade.php``` or ```inc/scripts.blade.php```, please make sure you [use the newest version of Bootstrap](https://github.com/Laravel-Backpack/Base/pull/362/files#diff-96ac3ea4d0cb85053acf44e3772eb5f1); they've fixed a security vulnerability (XSS);

red-ab:
Hi @tabacitu. I would like to update Backpack Base to 1.1.2, but some of my other packages still rely on Carbon 1, which blocks the install. Do you plan to integrate support for Carbon 1 instead of Carbon 2 only? Thanks.

tabacitu:
Hi @red-ab ,

Good question. I expect most packages to provide Carbon 2 support pretty soon. So _at this time_, **I don't think we should add support for Carbon 1, no**. Of course, I could change my mind if more people ask for Carbon 1. But it would lead to a little bit of spaghetti inside Backpack to pull it off, and I would like our code quality to go the other way :-)

What packages are you using that don't support Carbon 2 yet?

Cheers!

red-ab:
Hi @tabacitu. Unfortunately, those are packages I don't have any control on and which don't plan to integrate Carbon 2 support for the moment. It would have been great to have a Backpack Base version that integrates Laravel 5.8 support, but without the Carbon 2 breaking change for systems stuck with Carbon 1. Thanks for your support!

tabacitu:
I get it, thanks for the input @red-ab . If more people have this problem and want Carbon 1 support, please reply here so we know. If more people ask for it, we can add it.

Cheers!

PR #364, opened 2019-03-13
Shorter default date display and link to Carbon docs instead

ziming:
The reference table is under localization section of Carbon docs

https://carbon.nesbot.com/docs/#api-localization

I also shorten the date format from

```
'default_date_format'     => 'Do MMMM YYYY',
'default_datetime_format' => 'Do MMMM YYYY, HH:mm',
```

```
'default_date_format'     => 'D MMM YYYY',
'default_datetime_format' => 'D MMM YYYY, HH:mm',
```

Because 11 Sept 2019 is shorter than 11th September 2019. Which for admin views is almost always better given the number of table columns in a limited screen space. Allowing 1-3 more columns to show without hiding


tabacitu:
Definitely - thank you @ziming !

PR #376, opened 2019-05-15
[4.0][WIP] CoreUI Theme for Backpack

tabacitu:
Uses [CoreUI Free Admin Panel Theme](https://coreui.io/) instead of AdminLTE v2.
Uses Bootstrap 4.

Notes:
- Should probably be split into its own package after it's done;
- CoreUI CSS and JS are provided in the Packagist package, by copying the ```dist``` folder of CoreUI. There should be an alias from ```public/packages/backpack/coreui``` pointing to ```vendor/backpack/base/src/public/coreui```;

Todo:
- [x] layout views
- [x] auth views
- [x] my account views
- [x] crud views


tabacitu:
Wrong way to do it. Instead of having every freaking css js and html file from the CoreUI package inside our project, I'd rather we load them from CDN, and include the bare minimum as a fallback (primary css and js only).

Closing this as this PR https://github.com/Laravel-Backpack/Base/pull/380 should be a complete fix for this.

PR #377, opened 2019-05-15
[4.0][WIP] Tabler Theme

tabacitu:
After implementing the Tabler theme instead of AdminLTE 2, I'm not sure this is a good choice for us. It looks very good. Maybe the best-looking free admin theme. But:
- it uses require.js; it would make no sense for us to use require.js which seems on its way out; either we use webpack/mix, nothing at all, or our own solution; but not a technology that loads JS, but not also CSS, and that is used less and less these days;
- it has no sidebar; so either we create one, or live with only the top menu;
- the HTML isn't exactly clean; seeks like it needs a lot of divs to get the job done;

Screenshots:
Screenshot 2019-05-15 at 11 15 20
Screenshot 2019-05-15 at 11 16 06
Screenshot 2019-05-15 at 11 16 16
Screenshot 2019-05-15 at 11 16 31
Screenshot 2019-05-15 at 11 17 10
Screenshot 2019-05-15 at 11 17 19

All-in-all, I can't say I'm excited about moving to Tabler. Sure, it uses Bootstrap 4, and it looks good. But I don't feel it's as easy to work with as AdminLTE v2 was.

Maybe we should consider AdminLTE v3 and CoreUI as the default theme.


iMokhles:
personally i like top menu those days ( for me it looks good )

iMokhles:
and for choosing a them i suggest an installation command or option --theme='THEME_NAME' to publish the selected theme ( AdminLTE, CoreUI or Tabler ) to make developers life easier and of course with a default theme if developer forgot to type --theme='THEME_NAME'

imansugirman:
Amazing. 👍

johnpuddephatt:
It sure does look pretty.

Regarding your concerns about require.js... it looks to me like their JS mostly consists of jQuery, bootstrap.js and a few bits of code for making graphs/charts interactive.

Might not be a terrible idea to ignore their JS altogether and reimplement it with whatever build process you decided was best. There seems to be fairly little custom code.

tabacitu:
Thanks for the feedback guys. 

@johnpuddephatt indeed the only real option was to ignore their JS entirely and start from scratch. That plus the fact that the HTML wasn't exactly clean, often it used a different syntax than Boostrap (which... why would you?!), and the docs seem hasty, made me try the next options in the list, which were AdminLTE v3 (unreleased for more than a year but reasonably stable) and CoreUI.

In the end, after weeks of trying things out, I've decided it's better scrap everything we've done with Tabler and AdminLTE v3, and go with CoreUI instead. Here's the PR for it: https://github.com/Laravel-Backpack/Base/pull/380

It now looks similar, its usage is straightforward, it's better maintained than Tabler and it's actually released, as opposed to AdminLTE v3. Plus, its business model is simple - this is a free HTML theme, if you want themes with more components and different designs, you can purchase a premium one from the same supplier. I like that, I hope they come up with some good premium designs and people will support them that way.

PR #380, opened 2019-06-21
[4.0][Feature][Needs cleanup and polish] Use CoreUI and Bootstrap 4 by default

tabacitu:
Similar design to the 3.4 revamp, but using CoreUI instead of AdminLTE. I found this CoreUI was the only free Bootstrap 4 Admin Panel HTML Theme that was easy to work with, and provided a reasonable amount of building blocks (basically all of Bootstrap's plus a sidebar). And trust me, it took MONTHS to take this decision. In the process, I've tried the top 5 free admin panel themes, and even got to implement like 80% of the features for 2 of them (AdminLTE v3 and Tabler). But they each had stuff that annoyed me. I consider the HTML theme you're working with very _very_ important in admin panels, so I decided to scrap weeks of work and start again. Fortunately CoreUI is a keeper, in my opinion.

However, I found its default design underwhelming, so I went ahead and layered my design preferences on top of CoreUI. The result is [Backstrap](https://github.com/DigitallyHappy/BackStrap), a free HTML template that I hope more people will use, even outside of Laravel.

Screenshot 2019-06-21 at 11 49 34
Screenshot 2019-06-21 at 11 50 12
Screenshot 2019-06-21 at 11 50 01

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,11 +47,12 @@ php artisan backpack:base:install
 
 3) [optional] Change values in config/backpack/base.php to make the admin panel your own. Change menu color, project name, developer name etc.
 
-## Upgrading from Laravel 5.7 to Laravel 5.8 (or from Base 1.0 to Base 1.1)
-- Upgrade to Laravel 5.8; you might need to change your ```backpack/crud``` dependency to ```3.6.*``` in your ```composer.json```;
-- in your ```App\Models\BackpackUser``` instead of ```Tightenco\Parental\HasParent```, please use ```Backpack\Base\app\Models\Traits\InheritsRelationsFromParentModel```; [here's the diff](https://github.com/Laravel-Backpack/Base/pull/362/files#diff-f075b83ebb2b1ef3ba84dec14b395607);
-- in your ```app/config/backpack/base.php``` please change your ```default_date_format``` and ```default_datetime_format``` to ```Do MMMM YYYY``` and ```Do MMMM YYYY, HH:mm``` respectively;
-- if you've overwritten ```inc/head.blade.php``` or ```inc/scripts.blade.php```, please make sure you [use the newest version of Bootstrap](https://github.com/Laravel-Backpack/Base/pull/362/files#diff-96ac3ea4d0cb85053acf44e3772eb5f1); they've fixed a security vulnerability (XSS);
+## TODO: Upgrading from Base 1.1 to Base 2.0
+- Make sure you're running Base 1.1 and CRUD 3.6; If not, follow the upgrade guides to reach those versions;
+- Upgrade to Laravel 6.0; 
+- Change your ```backpack/crud``` dependency to ```4.*``` in your ```composer.json```;
+- more
+- even more
 
 
 ## Usage 


johnpuddephatt:
Great! CoreUI isn't *too* dissimilar to AdminLTE visually so hopefully it won't prove controversial.

Have only had a quick look for now but one quick question – is the Laravel-Backpack/HTML-Template package something that you're thinking would be maintained separately? Or will Laravel-Backpack/Base require it and pull in its compiled js and css?

johnpuddephatt:
Played around with this a bit more and I think with a bit of polish this has loads of potential!

Sharing screenshots rather than code to get quick feedback from people but very happy to implement.

## Login page
- Looks nice and clean!
- Might be more logical to move password reset to be next to password input, and expand the register text so it's a bit more informative?
- Could the 'powered by backpack' and 'handcrafted by ...' messages be be omitted here for simplicity?

Screenshot 2019-06-22 at 13 34 06

## Primary Navbar
- A white background and increased vertical padding would separate the navbar from the rest of the content without introducing too much contrast which would clutter the design.
- With a white background applied I don't think the nav-brand needs the reduced opacity, which was making it hard to read

## Sidebar
- This looks good. I like it without a background color, reduces clutter to my eye. Others will disagree so might it make sense to allow users to change the background color in the same way we currently change the AdminLTE theme color?
- In the current PR the nav items anchors are missing the class 'nav-item' to display them properly, I've added that in in my screenshot.

## Welcome message
- I've always thought the login status heading + 'you are logged in' message is a bit cluttered without communicating much. I wonder about changing this to a simple welcome message as pictured. Less cluttered so draws attention more to the sidebar.

## Breadcrumb
- The background on this was giving it more prominence than I think is necessary – I see this is a fallback method of navigation for more people (rather than primary or even secondary). Losing the background makes it less distracting, but I think a bottom border is probably necessary to separate it from the content that follows.

Screenshot 2019-06-22 at 14 25 47

## Account pages
- Introducing an additional navigation here looks messy, especially with it being vertical and in close proximity to the main sidebar.
- It's hard to lay these pages out well when the content is so short. Would it be better to combine the account details and password pages, removing the need for navigation? Each component could then be a card with a header and footer?

Screenshot 2019-06-23 at 13 16 20

johnpuddephatt:
Screenshot 2019-06-23 at 16 54 45

A bit more playing around: this time a GitHub inspired dark header, white sidebar and a lighter background color. I could see the header working well in other colours too, making the interface 'brandable' like AdminLTE.

Referencing other sites I'm realising the page title – e.g. My Account –  can probably be quite a bit smaller – having it too big distracts from other content.

As before this is definitely more for discussion rather than me saying this is the right approach!

tabacitu:
@johnpuddephatt thanks A LOT for the feedback. To answer your first question first: currently I think the best way to include the css & js is to have _only the compiled files_ in the Laravel-Backpack/Base repo. And publish them upon installation. This would fix another annoyance some people had with the way we installed AdminLTE - the fact that we had thousands of files inside /vendor/almasaeed2010/adminlte, that we then published. But the same thousand files, remained both in ```vendor``` and in ```public/vendor```. A bit messy. And it made some people hit their ceiling on free hosting. 

Yes, it would be a little more difficult to maintain - if we want to make any changes, we have to compile the file in the [BackpackUI](https://github.com/Laravel-Backpack/HTML-Template) repository, then overwrite that file in the Backpack\Base repository. But it would most likely only be one 1-3 files - the coreui css and js, and ```backpackui.css```. What do you think about this?

---

Your [Github-inspired design](https://github.com/Laravel-Backpack/Base/pull/380#issuecomment-504765309) here looks GREAT, by the way. Personally I would love to have that as an option in some projects. I've just pushed an update so that ```BackpackUI``` is an _overlay_ on top of CoreUI. I thinks this makes more sense. 

If you make the CSS file you made public as, people could put it in their ```public``` folder and use it really fast, instead of the ```backpackui.css``` file - just like they did in v3. 

So in ```config/base.php```:

```
    // Overlays - CSS files that change the look and feel of the admin panel
    'overlays' => [
        'packages/backpack/backpackui/css/backpackui.css', // default minimal look
        // 'packages/backpack/backpackui/css/github-inspired-ui.css', // Github-inspired look
    ],
```

That's how I though it would be easiest for people to customize the look-and-feel, with a plain and simple CSS file. What do you think? 

That's where the BackpackUI repository comes in handy. In order to create a custom overlay you can:
- [Install the BackpackUI repository](https://github.com/Laravel-Backpack/HTML-Template#installation); 
- run ```npm run serve``` to be able to browse through every design component;
- change anything you want inside the ```src/css/backpackui.css``` file to get the look you want;
- rename the ```src/css/backpackui.css``` file to something else, and use it in your Backpack project;
- (optionally) we could make it easy for people to minify it using ```npm run build```;

---

In terms of design changes to the backpackui css - I'll try them and get back to you. I agree with most of them :-) Thank you!

johnpuddephatt:
The approach you've described would work. It's quite simple and it solves a couple of issues.

### Some thoughts on themes...
An alternative (and I know it's already been discussed elsewhere) would be to have each 'theme' as a repository. The default theme could be installed by default (it would be listed as a dependency of Base so installed with it automatically), with other themes installable with composer. A config variable would set the theme, e.g. "backpack/default-theme", and this would tell Backpack where to load the files from (i.e. /vendor/backpack/default-theme).

#### Advantages
One advantage is that this could also handle custom views, making Backpack less tied in to one particular UI framework.

Another advantage is that you lose the thousands of files of bloat you mentioned. The package would come with compiled CSS and JS, just as you're proposing. Things like Bootstrap, CoreUI etc would *only* be present as separate files if the user wants to do development of the theme and runs `npm install`. Otherwise they just see the compiled files, and they don't have to do anything with npm/webpack/mix.

You would lose the reliance on CDNs with the issues that they can bring. And all dependencies would be under source control so things like upgrading a vendor library would be quick.

The above approach could work alongside overlays. Overlays could be positioned as a way to make quick customisations, while a theme is for more substantial changes.

#### A quick demo
I've [created a demo theme](https://github.com/johnpuddephatt/default-theme) to provide a simple examples of what one might look like – I've not integrated it into Backpack, but I tested out loading views from it by putting it in `vendor/johnpuddephatt/default-theme`, adding a `theme_name` config variable and adding the below code to BaseServiceProvider and it seemed to work.

````
        if (file_exists(realpath('../vendor/' . config('backpack.base.theme_name')))) {
            $this->loadViewsFrom(realpath('../vendor/'. config('backpack.base.theme_name') . '/views') , 'backpack');
        }
````

This bundles Bootstrap CSS, CoreUI's CSS and your custom backpack theme CSS into one single file. It means there's potential for nicely organised SCSS. 

The complied files would need to be published, but keeping them in the /public directory makes this easy to do with artisan's vendor:publish command.

Downloaded, the repo is ~600KB and contains 14 files, but if you clone the repo and open index.html you'll see all the necessary CSS and JS is there. Uncompressed there's around 300KB of CSS and 140KB of Javascript. Because the CSS file is a single file with a load of class name repetition between the sources it will compress very well – down to about ~40KB.

To develop the theme, you'd run npm install, make changes then run `npm prod` to create the production files.

Sorry for the long comment, probably not the best place to discuss this but thought it was a good time to share!

ridislam:
How can I use this theme to my existing backpack project? is there any manual to use it?

tabacitu:
@ridislam no, not yet. It will only be available in v4. Subscribe to the newsletter and we'll let you know when it's out. Or stay in touch with v4 development in this PR: https://github.com/Laravel-Backpack/CRUD/pull/1935

tabacitu:
Note: Merged into the ```v4``` branch - https://github.com/Laravel-Backpack/Base/pull/384 . Feedback and changes for these features should still happen on this feature branch, before merging to ```v4``` again.

tabacitu:
@johnpuddephatt , just implemented most of your design changes - again, thank you. I agree with you on most of the issues you've raised. New screenshots:

Screenshot 2019-08-01 at 17 43 10
Screenshot 2019-08-01 at 17 43 29

I think it's really starting to come together. I expect there will be a lot of small design changes after alpha is launched, when more people get to install it and play around.

tabacitu:
@johnpuddephatt on themes - totally agree with you they should be separate composer repositories. Your default-theme repo is great, thank you. It's just the way I imagined it. Might end up forking it for the default theme :-)

I think @AbbyJanke 's approach in [this PR for Base themes](https://github.com/Laravel-Backpack/Base/pull/297) (using helpers) provides a good solution. Let me know what you think there - it's the last major task for Base before we hit 4.0-alpha, so it's next on my agenda. I'll play around with different solutions to see what works best., but hers looks like the winner right now.

tabacitu:
This has been merged into the ```v4``` branch for the alpha release and it's working very well. ```v4``` even has a few more design improvements than this branch, [mentioned here](https://github.com/Laravel-Backpack/Base/pull/380#issuecomment-517319199).

This branch is now obsolete, let's follow ```v4```. Design improvements are welcome, but they will need to happen starting from the ```v4``` branch, not this one.

PR #382, opened 2019-07-19
Sync with tightenco/parental latest bug fixes

ziming:
That package got updated again with some fixes. This PR simply copy over the changes.

This trait now support 5.8 so you may load their trait or merge this if you think they won't be updated fast enough for the upcoming 5.9.


tabacitu:
You're totally right! Thank you for putting this together @ziming  - excellent PR. Merged and tagged, available as 1.1.7 with ```composer update```.

Thanks again! Cheers!

PR #381, opened 2019-06-24
[4.0][Feature][Ready] Ability to show/hide breadcrumbs

tabacitu:
Allows you to hide all breadcrumbs across the admin panel.
Or show them. Not sure which one should be the default :-)

Requires the CoreUI branch to be merged first.

Screenshot 2019-06-24 at 11 57 57
Screenshot 2019-06-24 at 11 58 12

PR #384, opened 2019-07-25
[4.0][Branch][WIP] Base for Backpack v4-alpha

tabacitu:
Features will be checked as they're merge into this branch:

### MUST

- [x] uses CoreUI - https://github.com/Laravel-Backpack/Base/pull/380
- [x] toggle breadcrumbs on/off - https://github.com/Laravel-Backpack/Base/pull/381 [ready]
- [x] Ability to load assets with or without Laravel Mix; https://github.com/Laravel-Backpack/Base/pull/385 [ready]
- [x] Ability to easily use Backpack without any CDNs; https://github.com/Laravel-Backpack/Base/pull/385 [ready]
- [x] Easily enable Vue.js in your admin panel; https://github.com/Laravel-Backpack/Base/pull/385 [ready]
- [x] Ability to show/hide breadcrumbs; Laravel-Backpack/Base#381 [ready, docs updated]
- [x] API to change breadcrumb links from Controllers; https://github.com/Laravel-Backpack/Base/pull/389 [ready, docs updated]
- [x] themes - https://github.com/Laravel-Backpack/Base/pull/297;
- [x] Widgets / Cards; easily add stats, pie charts, line charts and other quick stats to a dashboard or CRUD (or any part of your admin panel); https://github.com/Laravel-Backpack/Base/pull/390
- [x] SweetAlerts; instead of showing the default browser alert(), Backpack now uses much prettier SweetAlert pop-ups; custom design included;
- [x] views are now split into a lot of small files, so you can easily overwrite just a tiny bit;
- [ ] fix lang file naming, each Backpack package should have its own namespace; 

### SHOULD

- [x] Installation - one package instead of two;
- [x] Use webfactor/generators instead of backpack/generators and laracasts/generators
- [ ] Base - Installation wizard now asks if you want a panel to manage your admins/users and installs UserManager or PermissionManager;

### COULD

- [ ] Authentication - it’s now a separate package;
- [ ] SocialAuthentication - adds Social login to your admin panel, if you would like your admins to login via Facebook, Google, Github, LinkedIn, etc;

### WON'T

- [x] Backpack v4 won't be a SPA; we'd like to move in that direction, but in my opinion, right now the CONs outweigh the PROs; routes are written twice, you have both blade templates and JS templates, NOT exposing sensitive information in JS is an issue, learning curve is steep for juniors and mid developers, [pages are not always faster to load](https://carter.sande.duodecima.technology/javascript-page-navigation/) and [browsers themselves recommend using server-side rendering in 2019, not client-side rendering](https://developers.google.com/web/updates/2019/02/rendering-on-the-web); honestly - I'm not happy with how SPAs _can_ be built right now; I just don't think the JS ecosystem is mature & stable enough to use _by freelancers and agencies_, for client projects that you have to maintain for years; I _do_ think this will most likely change when [Inertia.JS](https://reinink.ca/articles/introducing-inertia-js) or [Livewire](https://livewire-framework.com) are launched and stable (and maybe even Vue v3); I _can't wait_ to finally have the front-end completely separated from the back-end; but until SPA technology is good enough to recommend for _most projects_, server-side rendering is _still_ our preferred way of building admin panels; and I think it should be for most freelancers, agencies and software companies too;


tabacitu:
All Backpack/Base files have been moved into Backpack/CRUD, when we launched Backpack v4.0. 
So this PR is only needed to highlight the differences between v3.6 and v4.0 - which can be done if it's closed too.

PR #385, opened 2019-07-26
[4.0][Feature][Ready] Use Laravel Mix. Remove CDNs.

tabacitu:
This branch:
- uses NPM to pull the assets that Base needs;
- uses Laravel Mix to create a ```bundle.css``` and a ```bundle.js``` file;
- uses Laravel Mix to copy the above bundle files, plus the font files, to the package's ```public``` folder;
- use Laravel Mix to copy the css&js assets from the package's ```public``` folder to the app's ```public``` folder (this will only be run by Backpack maintainers);

In order to install/update dependencies, the npm commands need to be run inside the ```vendor/backpack/base``` folder:
- ```npm install```
- ```npm run prod```

----

The end result:

Screenshot 2019-07-27 at 09 57 47

----

Notes:
1) By default Backpack/Base now loads the bundle CSS and JS files, not individual files for jQuery, Bootstrap, Coreui, Source Sans Pro, PNotify, Pace.js, etc. 
 - PROs: Smaller file size in total. One request instead of many;
 - CONs: More difficult for developer to customize (overwrite blade file, replace with individual files from CDN/local)
2) The bundle files are pretty big (289KB ```bundle.css```, 329KB ```bundle.js```). But this does inlcude jQuery, Bootstrap, CoreUI and the Source Sans Pro font family, so that Backpack\Base is completely CDN-free. Better performance can most likely be achieved by using CDNs for jQuery, Bootstrap & font, and leveraging HTTP/2 concurrent requests. But then it's no longer CDN-free. So the tradeoff here is uptime-security-GDPR over performance.
3) I've chosen to load the Backstrap overlay _outside_ the bundle. So that people can easily remove or overwrite our opinionated design, by removing the overlay file from ```config/backpack/base.php``` and/or adding their own.
4) Why use ```asset()``` with our own cachebusting (current Backpack/Base version), instead of the ```mix()``` helper? It allows us (maintainers) to use Laravel Mix, even though the end-user might not.

Feedback? Let me know what you think.


tabacitu:
Note: Merged into the ```v4``` branch - https://github.com/Laravel-Backpack/Base/pull/384 . Feedback and changes for these features should still happen on this feature branch, before merging to ```v4``` again.

marconett:
amazing, can't wait to use this. keep it up 🙏

PR #387, opened 2019-07-27
[4.0][Feature][Ready] Ability to define JS&CSS assets in config file

### LINKED ISSUE #386: [4.0][Feature][Proposal] Easily enable Vue.JS or React
Sometimes people want to use Vue.JS or React inside their Backpack admin panels. Or other javascript libraries. This [was possible in v3 too](https://medium.com/@owenmelbourne/using-backpack-for-laravel-with-vuejs-or-even-react-b6e4ea48caf7), but in v4 we could make it even easier.

What do you think about allowing people to add stuff to an array in ```config/backpack/base.php``` when they want css/js added to all pages? We could have:

```php
    // ---------------
    // CSS & JS ASSETS
    // ---------------
    // CSS & JS files that are loaded in all pages, using Laravel's asset() helper 
    
    'styles' => [
        'packages/backpack/base/css/bundle.css', // includes jQuery, Bootstrap, CoreUI, PNotify, Popper
        'packages/@digitallyhappy/backstrap/css/overlay.css', // default design on top of CoreUI
    ],

    'scripts' => [
        'packages/backpack/base/js/bundle.js', // includes jQuery, Bootstrap, CoreUI, PNotify, Popper
        // examples
        // 'https://unpkg.com/vue@2.4.4/dist/vue.min.js',
        // 'https://unpkg.com/react@16/umd/react.production.min.js',
        // 'https://unpkg.com/react-dom@16/umd/react-dom.production.min.js',
    ],

    // ------------------
    // LARAVEL MIX ASSETS
    // ------------------
    // CSS & JS files that are loaded in all pages, using Laravel's mix() helper 

    'mix_styles' => [
        // 'css/app.css',
    ],

    'mix_scripts' => [
       //  'js/app.js',
    ],

```

We already had something similar with AdminLTE "_overlays_", allowing people to add CSS on top of AdminLTE, to customize the looks. The above would just add the possibility to do the same for JS, and optionally using Laravel Mix.

@OwenMelbz what do you think?

tabacitu:
Fixes #386 
Depends on https://github.com/Laravel-Backpack/Base/pull/385 which depends on https://github.com/Laravel-Backpack/Base/pull/380

Makes it super-easy to use VueJS - just uncomment one line in the config file.
Similarly, super-easy to use CDNs for Base assets if you prefer that.

All Base CSS&JS assets are now loaded from a single source of truth - the ```config/backpack/base.php``` file. I've included four arrays:
- ```styles``` and ```scripts``` - loaded with the ```asset()``` helper;
- ```mix_styles``` and ```mix_scripts``` - loaded with the ```mix()``` helper;

For a demo/example [take a look at the base file](https://github.com/Laravel-Backpack/Base/pull/387/files#diff-41252036ab915c73d727350733dde754) - an example will be included there in comments.


tabacitu:
Merged into v4. Added to docs. Closing this.

PR #405, opened 2019-09-04
inc.topbar_right_content was not loaded correctly

trakons:
`@include('backpack::inc.topbar_right_content')` was not loaded when `config('backpack.base.setup_auth_routes')` is `false`


tabacitu:
You're right, of course. Thank you for submitting the PR @trakons ! It's now merged and tagged, a ```composer update``` should do it for you. Cheers!

PR #413, opened 2019-10-04
Adds a configurable middleware group for the base routes

MartijnvAdrichem:
This pull request adds the functionality to make the middleware of the default routes configurable. In my case this is needed because my web middleware is already used by an other part of the application.

### DIFF
--- a/src/config/backpack/base.php
+++ b/src/config/backpack/base.php
@@ -68,6 +68,9 @@ return [
     // You can make sure all your URLs use this prefix by using the backpack_url() helper instead of url()
     'route_prefix' => 'admin',
 
+    // The middleware (group) used in all base routes
+    'route_middleware' => 'web',
+
     // Set this to false if you would like to use your own AuthController and PasswordController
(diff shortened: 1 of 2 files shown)


tabacitu:
Hi @MartijnvAdrichem ,

Thank you for taking the time to submit the PR. It's much appreciated. And sorry for taking so long to look at this - I kept postponing it since it wasn't an urgent matter.

I think it's a great idea! However 😄 This package (Backpack\Base) is no longer used by Backpack v4, and we're no longer pushing new features to Backpack v3 (which _does_ use this package). Starting with Backpack v4 everything Base does is included in Backpack\CRUD. One package to rule them all, if you will.

Please let me know if you want this feature in v4 too. If you do, let me know if you can submit the same PR, but for https://github.com/laravel-backpack/crud instead of backpack/base. I'd love it if you could do so, that way you'll show up as the contributor of this feature.

Thanks again - and sorry for the long wait.
Cheers!

PR #420, opened 2019-12-30
Overwrite loggedOut instead of logout in LoginController

lj3lj3:
In laravel 5.6.27, there is a `loggedOut` function been added in the `LoginController` which for the logics those after user logout, and since we depend on the laravel version higher or equal than 5.8, so it's safe to overwrite `loggedOut` instead of `logout` function.

### DIFF
--- a/src/app/Http/Controllers/Auth/LoginController.php
+++ b/src/app/Http/Controllers/Auth/LoginController.php
@@ -63,18 +63,13 @@ class LoginController extends Controller
     }
 
     /**
-     * Log the user out and redirect him to specific location.
+     * The user has logged out of the application.
      *
-     * @param \Illuminate\Http\Request $request
-     *
-     * @return \Illuminate\Http\Response
+     * @param  \Illuminate\Http\Request $request
+     * @return mixed
      */
-    public function logout(Request $request)
+    protected function loggedOut(Request $request)
     {
-        // Do the default logout procedure
-        $this->guard()->logout();
-
-        // And redirect to custom location
         return redirect($this->redirectAfterLogout);
     }
 
--- a/src/app/Http/Controllers/Auth/LoginController.php
+++ b/src/app/Http/Controllers/Auth/LoginController.php
@@ -66,6 +66,7 @@ class LoginController extends Controller
      * The user has logged out of the application.
      *
      * @param \Illuminate\Http\Request $request
+     *
      * @return mixed
      */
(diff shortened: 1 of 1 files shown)


pxpm:
Hello @lj3lj3 
Thanks for taking time to put this PR together.
I think the `loggedOut` function is when user already logged out of the APP, our logout function is the endpoint to log the user out. 

From Laravel docs: 
```
protected mixed loggedOut(Request $request)
The user has logged out of the application.

```
I think it's not replaceable by i migth be wrong. 

Best,
Pedro

lj3lj3:
Hi @pxpm, thank you for responding to my PR.

First of all, you're right. Overwriting `logout` function is more properly when it needs to be an endpoint to DO the logout staff, but what could happen is that when Laravel changes some logics in the `logout` function, we can't keep it up until we manually update it, like this git commit [698a6356b2aa91cd61f07832330832dcef7d7883](https://github.com/laravel/framework/commit/698a6356b2aa91cd61f07832330832dcef7d7883). 

So what I trying to say is that if we have some further plans to overwrite the default logics in the `logout` function, then we're cool, nothing needs to be changed, but if we just want to change the default redirect URL, then overwriting the `loggedOut` function seems a more reasonable idea to me, 'cause it is a kind of following laravel's way.

Please let me know what you think.

Good day!

tabacitu:
Hi guys,

I think you're both right. Thanks for noticing @lj3lj3 - I didn't know ```loggedOut()``` is now triggered immediately after ```logout()```. **I agree with the change, let's do it.**

No, we don't have any plans to _overwrite the default logics in the logout function_. Our change was just to overwrite _where_ the user will be redirected. We overwrote ```logout()``` because that's how the stock Laravel controller worked at the time, when we made the change ```loggedOut()``` didn't exist :-) Now that ```logout()``` does ```return $this->loggedOut($request) ?: redirect('/');``` it's better for us to overwrite ```loggedOut()```, I agree.

However, this package (Backpack\Base) is now deprecated. This package is only used by Backpack 3.6 and below, and we don't plan to make any more changes to it - unless it's a security issue. Everything inside this package has been moved to [Backpack\CRUD](https://github.com/laravel-backpack/crud/) in version 4, it's all under one roof. **Can you please create the same PR for that repo too?** So that you show up as the rightful author of this for Backpack v4+?

Thanks again.
Cheers!

PR #427, opened 2020-03-18
Fix Password Issue

ebrahimahmadi:
Hash::Make() will create wrong has from a password which leads to unauthorized error after password updates.

### DIFF
--- a/src/app/Http/Controllers/Auth/MyAccountController.php
+++ b/src/app/Http/Controllers/Auth/MyAccountController.php
@@ -61,7 +61,7 @@ class MyAccountController extends Controller
     public function postChangePasswordForm(ChangePasswordRequest $request)
     {
         $user = $this->guard()->user();
-        $user->password = Hash::make($request->new_password);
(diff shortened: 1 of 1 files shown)


tabacitu:
Hi @ebrahimahmadi ,

Are you getting this error in Backpack 3.x or 4.x? 
What's the output of your ```php artisan backpack:version```?

Cheers!

ebrahimahmadi:
> php artisan backpack:version

4.0.56

tabacitu:
Hmm... In that case @ebrahimahmadi , you're not actually using Backpack\Base at all. Starting with 4.0.0 all this functionality is inside Backpack\CRUD. So I'll move the issue there.

But. I've just tested this, an I've been able to successfully change my password, then login with the new one:


So it doesn't look like there's a bug at all. Please correct me if this is not the thing you were trying to correct.


I'm going to close the issue, since I'm not able to replicate it. Please come back with more information to help us understand how to replicate the issue, or what your issue is if it's not the above, and we'll reopen.

Cheers!

PR #431, opened 2020-07-06
Missing warning

fellwell5:
translated warning
removed dot after unauthorized because its one word


tabacitu:
Thank you @fellwell5 . Merging now. Keep in mind that Backpack/Base is no longer actively maintained, officially. Backpack/Base is only used by Backpack 3.x and we're now at Backpack 4.1 already (two version ahead). 

So if you're using Backpack 4.x please consider making the same PR towards Backpack/CRUD - that's where all this code lives nowadays.

Cheers! Thanks again.

PR #435, opened 2020-10-13
Add Support for latest laravel versions (7.x, 8.x)

lamungu:
This PR's purpose is to add compatibility for `backpack/base` onto latest releases of Laravel 7.x and 8.x

Change involves `composer.json`'s package list, akin to the last PR involving a Laravel upgrade  (#403)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     ],
     "require": {
         "almasaeed2010/adminlte" : "2.4.*",
-        "laravel/framework": "5.8.*|^6.0",
+        "laravel/framework": "5.8.*|^6.0|^7.0|^8.0",
         "prologue/alerts": "^0.4.1",
         "creativeorange/gravatar": "~1.0",
         "nesbot/carbon": "^2.14.0",


tabacitu:
Hi @lamungu ,

Thanks for the PR, but I'm pretty sure it's not that easy. In those versions Laravel has moved its Authentication to a different packages + a number of other breaking changes that would affect this package. And it will continue to do so, I'm certain.

TLDR:
- **Backpack 4.x already offers support for Laravel 7.x and 8.x**; and it does not use Backpack\Base at all; for using Backpack with Laravel 7.x, 8.x and above - please use Backpack 4.x;
- Backpack 3.x (which requires Backpack\Base) will not be receiving support for future versions of Laravel past 6.x, it's been superseded by Backpack 4.x (released more than one year ago - Sep 2019); and it's deprecated; it's not archived on Github just so that people who are unable to upgrade can still use it; but people really _should_ be upgrading their apps;

Check out https://github.com/Laravel-Backpack/CRUD 

Cheers!

tabacitu:
PS. I've just updated the README so that the above is clearer. Sorry for that - looks like it wasn't clear at all. I've also marked the package as Abandoned on Packagist - hopefully this will nudge people into upgrading to 4.x

If this causes any problems to anyone, please let me know!

